PR #321, opened 2013-10-29
DomSceneObject class was added with corresponding example

PixelsCommander:
DomSceneObject is a class that allows to stick HTML content to specific project point. It can be useful for commenting scene and enrich it with full variety of JS DOM components .

### DIFF
--- a/examples/DomSceneObject/Comments.html
+++ b/examples/DomSceneObject/Comments.html
@@ -1,0 +1,73 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title></title>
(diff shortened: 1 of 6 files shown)


lehni:
http://teempla.com/ looks great! Is paper.js still in use on it? I think as part of #743 it would be good to finally support HTMLItem. And ForeignItem wasn't such a weird suggestion after all, since in SVG it's called < foreignObject> : )

lehni:
I'm still interested in this. Just didn't find the time yet, sorry!

PR #335, opened 2013-11-06
How JS is compatible with IE8 ?

lencetang:
How JS is compatible with IE8 ?


bebensiganteng:
Was just wondering about this, how would one gracefully fall back when canvas is not supported.

In IE8 paper.js just froze everything up.

PR #582, opened 2014-12-02
First working version of paperjs with Worker detection

louisremi:
This PR replaces #581 
More tests are needed before it can be merged.


lehni:
There's a separate issue discussing this also: #634

lehni:
@louisremi thanks to the recent work on creating a unified version for both Node.js and the browsers, I managed to finally implement this now. Please test with the latest version and let me know how this works for you. There's also an example illustrating how to use it here:
https://github.com/paperjs/paper.js/tree/develop/examples/Worker

oncletom:
That's ace! 👏

PR #586, opened 2014-12-04
Resize canvas fix

### LINKED ISSUE #580: Resizable canvas doesn't size correctly when offset from right of viewport
The cause of this issue can be found here: https://github.com/paperjs/paper.js/blob/master/src/view/View.js#L60-63

Say I set a canvas to `100%` width and resize, but it's parent container is only 800px. If the canvas is set to the right of the viewport this works because `var offset = DomElement.getOffset(element, true)` refers to the left offset of the element.

Unfortunately it never subtracts the right offset.

[Example with left working correctly](http://jsfiddle.net/bLg9qdLf/5/)
[Example with right working incorrectly](http://jsfiddle.net/zwh8s0ns/4/)

In both the examples above note the center position of the circle, and you can also open up the console and I log out the size of the body element and the canvas element.

georeith:
Fixes #580

Resizable views were only taking into account their offset from the top left of the viewport.

Now when a view is resizable it doesn't explicitly set the canvas dimensional style properties, instead it just recalculates the coordinate space by setting the width and height attributes and allows the user's styling to define how the canvas resizes.

Broken example in the issue: http://jsfiddle.net/zwh8s0ns/4/
Same example with new library: http://jsfiddle.net/no3udkou/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+build/zip
--- a/dist/paper-core.js
+++ b/dist/paper-core.js
@@ -9,7 +9,7 @@
  *
  * All rights reserved.
  *
- * Date: Sat Nov 22 09:01:01 2014 -0800
+ * Date: Tue Dec 2 22:31:20 2014 -0800
  *
  ***
  *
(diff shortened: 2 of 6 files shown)


georeith:
@lehni  Thanks for doing the work. I was away over the holidays so didn't get a chance to do this, apologies.

Issue #461, opened 2014-04-27
Browserify/CommonJS export


jbielick:
I recently started using Browserify in a project using paper.js. If you're not familiar, browserify builds a bundle file of all the scripts found in a dependency tree defined in require('../module') calls.

In most cases, this means using npm to install packages locally, and when browserify builds the dep tree, modules are easily included via their resolved paths in node_modules.

However, since paper.js is available through bower and is also commonJS ready (I think?), it looks like there might be an issue with requiring paper js as a commonJS module from a bower directory.

If `bower install -S paper` is used, I can then require the paperjs module in my application with `require('../bower_components/paper/dist/paper-full')` (an index.js file would make this cleaner). However, when requiring the paper-full.js file in that manner, https://github.com/paperjs/paper.js/blob/master/dist/paper-full.js#L227 returns Base as the export, which turns out to be an empty array in this instance.

I guess this line https://github.com/paperjs/paper.js/blob/master/src/export.js#L51 gets lost in the build when it isn't a node env?, but it isn't at the end of paper-full.js. An AMD module export _is_ defined at the end and it seems like `module.exports = paper` could also appear in the `__options.environment == 'browser'` closure.

PR #589, opened 2014-12-09
Export the browser version when using browserify

megawac:
Currently browserify builds in the node version (could you do a patch release svp on npm)

See #461

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,9 +13,11 @@
     "Jonathan Puckey <jonathan@studiomoniker.com> (http://studiomoniker.com)"
   ],
   "main": "dist/paper-node.js",
+  "browser": "dist/paper-full.js",
   "files": [
     "AUTHORS.md",
     "dist/paper-node.js",
+    "dist/paper-full.js",
     "examples/Node.js",
     "LICENSE.txt",
     "README.md"
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "AUTHORS.md",
     "dist/paper-node.js",
     "dist/paper-full.js",
+    "dist/paper-core.js",
     "examples/Node.js",
     "LICENSE.txt",
     "README.md"


chad-autry:
I think a better long term solution would be to actually publish 2 different modules in npm. For me on windows some of the dependencies prevent me from running npm install of the existing module.

My temporary solution at the moment was to fork paper.js, edit the dependencies, and retrieve the project directly from github instead of npm.

lehni:
Yeah this needs addressing... Is there any way to change the dependencies for browser? Or is two separate modules the only way to go?

lehni:
@chad-autry which dependencies cause problems on windows? I imagine it's node-canvas?

megawac:
Optionaldependency it?

chad-autry:
> canvas@1.2.1 install \node_modules\paper\node_mod
ules\canvas
> node-gyp rebuild
gyp ERR! stack Error: Can't find Python executable "python", you can set the PYT
HON env variable.

I suppose more technically I should say it is a lack of Python, and not directly Windows at fault.

There isn't a way to drive dependencies off the install type that I could see (the "browser" element megawac used is part of browserify). As megawac says, an optional dependency could be added, but then someone who actually wants the node.js version might not realize they're missing a dependency.


I think the best solution is to publish two different versions to npm under different package names. Can do it fairly easily and manually from a second repo (update only /dist/paper-core.js when the main repo is updated and republish). If desired the main repo's build scripts could eventually be updated to publish both versions.

lehni:
Yeah I think you're right, two separate modules appears to be the way to go. But people will still require the wrong one then, most of the time :)

chad-autry:
FYI this fork of mine is available for getting just paper-core.js through npm for use with browserify. https://github.com/chad-autry/paper.js I haven't published this to npm yet, though since npm can retrieve modules straight from github it is perfectly usable as is.

That is something to think about for the two repo approach vs adding double publishing into build scripts. With two github repos they are both install-able straight from github in case there is an issue with the npm repo. If both versions were published from the same repo that might break npm installation from github.

I tried re-tagging v0.9.22, but I don't seem to be able to override the tag that came when I forked.

lehni:
I am going to merge this request either way, and then think about the creation of a separate browser based npm package some more.

lehni:
I have merged this into `develop`. It will make it into `master` with the next release. Pull requests should be sent to the `develop` branch from now on.

lehni:
We could have two separate NPM packages in addition to `paper`: `paper-core` and `paper-full`, or perhaps `paper-browserify-core` and `paper-browserify-full`. These could be branches with the same name, with modified package.json files (and perhaps README too), that get published on each release through `npm publish`.

chad-autry:
The branches solution sounds good to me, I'd make use of it.

PR #591, opened 2014-12-13
Browser detection: Add Atom shell

ponychicken:
I'm developing a Atom plugin that will add a paper.js repl. Eg. you can edit your paper code in one panel and see the visual result in another panel.

Atom is running in an unsual mix of both node and browser environment at the same time. Currently paper.browser detects it as browser.chrome. This enables some feature that aren't helpful in atom, so with this commit I added another browser to paper.browser.


lehni:
@ponychicken great work on that plugin! I just tried it out. Is there a way to change the canvas size?

ponychicken:
Right now there isn't (it only resizes when the the window resizes). Feel free to open a bug: https://github.com/ponychicken/atom-paper.js

lehni:
We could consider a way for the script to change the view size, and your plugin respond to that by resizing itself?

Issue #592, opened 2014-12-13
Don't shim ES5 methods


megawac:
If a user cares about old, old firefox browsers which have canvas they can include the ES5 shim. Its time :)

According to caniuse + kangax tables, the browsers which support `canvas` but don't have `isArray` account for `0.07%` of the market

That may or may not also extend to the `Http` shim, depending if you care about supporting svg canvas shims

PR #593, opened 2014-12-13
Helpers

megawac:
Related to #592, toying around with the lib this morning, this pr offers several simplifications (also will be faster) dependent on `Object.keys`. An alternative is to 1) shim keys here or in straps if #592 is punted 2) reject :)


lehni:
Now that I've merged #594, this one makes sense too.

lehni:
Merged in 8f9e0f58fa7b9b1a161e76063b9515329d90d72d.

PR #594, opened 2014-12-13
Refactor and optimize base equals

megawac:
Do the `checkKeys` pass via `Object.keys` (if they have the same num of keys then the next pass will clearly catch all vallues), also reuse some of the array code with the obj check.

This is more or less a port of underscores is equal check for objs and arrays.

Note currently types aren't checked so `paper.Base.equals(new paper.Point, {y: 0, x:0})` which may or may not make sense.


lehni:
This is merged in 4f17aa376d3694f96ca06a39d01375359b455338. Thanks and apologies for taking so long to review!

PR #595, opened 2014-12-13
Reuse the Tool mousedrag handling logic for mousemove

megawac:
I'm not sure how I personally feel about my own refactor lol, so close if its not clear. Reuses a lot of the logic that is shared between the event handling for `mousemove` and `mousedrag`. Comments should be updated, didn't get around to it yet

(this is quite similar to how you are doing in in CanvasView)

### DIFF
--- a/src/tool/Tool.js
+++ b/src/tool/Tool.js
@@ -354,29 +354,11 @@ var Tool = PaperScopeItem.extend(/** @lends Tool# */{
         // Update global reference to this scope.
         paper = this._scope;
         // Now handle event callbacks
-        var called = false;
+        var called = false, drag = false;
         switch (type) {
         case 'mousedown':
             this._updateEvent(type, point, null, null, true, false, false);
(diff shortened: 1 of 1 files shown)


lehni:
Could you explain a bit why you needed the same logic for moves as for drags? The current implementation respects min / max distance only when actually using the tool (dragging, so to speak). mousemove is meant to be a more low level event that is fired when the mouse isn't dragging, and also during drags when now mousedrag handler is defined.

lehni:
I didn't look closely enough at the code when writing the above. Looking again, this sure makes a lot of sense, and you were right that this is a nice simplification. But there is one catch: You've built in an endless loop in mousemove. I'll merge this in and then fix that loop.

lehni:
This is now merged, but since I had to merge against `develop`, Github doesn't seem to have detected that. Thanks for all the pull requests! Nice work, would be great if you found more time to work on paper.js!

PR #600, opened 2014-12-16
Update FutureSplash.html

Patak123:
Hi :)
segment.previous.fixed = Undefined
segment.previous.point.fixed = True / False


lehni:
You're right. Thanks!

PR #624, opened 2015-01-23
Add .interpolate() method to Segment, Path and CompoundPath

louisremi:
Basic interpolation/extrapolation between two existing items.

I'm not sure how `this._changed()` is supposed to work in `Segment` and `Path`, please make sure I handled it correctly.

### DIFF
--- a/dist/paper-core.js
+++ b/dist/paper-core.js
@@ -9,7 +9,7 @@
  *
  * All rights reserved.
  *
- * Date: Sun Oct 25 11:23:38 2015 +0100
+ * Date: Sun Oct 25 11:24:42 2015 +0100
  *
  ***
  *
@@ -88,7 +88,7 @@ var Base = new function() {
 				val = dest[val.substring(1)] || val;
 			var isFunc = typeof val === 'function',
 				res = val,
-				prev = preserve || isFunc && !val.base
+				prev = preserve || isFunc
 						? (val && val.get ? name in dest : dest[name])
 						: null,
 				bean;
@@ -136,7 +136,7 @@ var Base = new function() {
 
 	function set(obj, props, exclude) {
 		for (var key in props)
-			if (props.hasOwnProperty(key) && !(exclude && exclude[key]))
+			if (props.hasOwnProperty(key) && (!exclude || !exclude[key]))
 				obj[key] = props[key];
 		return obj;
 	}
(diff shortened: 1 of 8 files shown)


lehni:
It looks good! I am wondering if we should define the behavior for when the segment count is not the same?

louisremi:
The interpolation is basic : it stops interpolating when the last segment of the item with the smallest number of children has been interpolated. It doesn't handle interpolation between a CompoundPath and a Path, I'm not sure what the use-case would be.

For more complex interpolation, we should consider porting [MutatorMath](https://github.com/LettError/MutatorMath/) to JS and the Paper API. I'm not sure but it might be able to handle different number of children in an smart way.

timuric:
@louisremi @louisremi Why not to use path.flatten() before interpolating to make sure that path contain equal amount of segments?

louisremi:
> flatten(maxDistance) 
> Converts the curves in a path to straight lines with an even distribution of points. The distance between the produced segments is as close as possible to the value specified by the maxDistance parameter.

Doesn't sound like what I'm trying to achieve here. But maybe I misunderstood your proposition.

timuric:
@louisremi yes, I see it now. I have been doing interpolation of complex paths where amount of points vary drastically, and basically I end up doing something like this:
`pathA.segments.map((x) => x.point.between(pathB.getNearestPoint(x.point))) `
however this turned out to be slow, and as a dirty workaround I end up using path.flatten() to equalise amount of segments and interpolating between them without using getNearestPoint method. However I worked only with hand drawn paths, which are quite different from geometric shapes or bezier paths.

lehni:
I have cherry picked your commit above into the https://github.com/paperjs/paper.js/tree/byte-foundry-interpolate branch. I will work a bit on it there, and then eventually merge it in, finally.

The cherry-pick was necessary for a few reasons:

- Your pull request is against `master`, not `develop`. Please note that all PRs should now be against `develop`
- You included built library files in the PR. Those should always be excluded, only the source files that changed should be included.

More on this soon!

lehni:
BTW, did you test on CompoundPaths? Doesn't look like this would work?

https://github.com/byte-foundry/paper.js/commit/2d0a9d53bb59d281a7eb4f5dcebccda9352b737a?diff=unified#diff-9ce5bdb8a22ef78461d619b28d1d3e17R8824

The parameters are called `compoundpath0/1`, but the variables in the loop `path0/1`...

lehni:
@louisremi I hope you're happy with how I implemented it. The path / compound path on which you call interpolate can be empty and is filled accordingly, the operands need to have the same amount of children / segments.

PR #627, opened 2015-01-26
Issue 626: Don't try to iterate a null array

iam-TJ:
Whilst using ImportSVG() some SVG files were causing a premature failure resulting in the paper.js code aborting. I tracked it down to a potential null array in setPathData() if the data.match(...) didn't produce any matches.

### DIFF
--- a/src/path/PathItem.js
+++ b/src/path/PathItem.js
@@ -163,6 +163,8 @@ var PathItem = Item.extend(/** @lends PathItem# */{
             current = new Point(),
             start = new Point();
 
+        if (!parts) return;
+
         function getCoord(index, coord) {
             var val = +coords[index];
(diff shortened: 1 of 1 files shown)


lehni:
Could you provide the path data string that caused this issue? Thanks!

iam-TJ:
It occurred during the global game jam so I've not looked at the code since. I think the issue was caused by an AI or EPS file converted to SVG by Inkscape. I *think* the file was "earth-cartoon.svg" which can be found at

https://github.com/iam-TJ/ggj/tree/54cc2ffc6073dabebf1c47c3b24d9f19449a61be/src/assets

lehni:
Yes indeed, that file is causing it. Thanks!

Issue #642, opened 2015-02-23
Symbol's stroke clipped with exportSVG 


alexburner:
I've noticed that a Path.Circle with a stroke exports to SVG fine, but when I place the Circle as a Symbol, the exported SVG clips the stroke to the shape's bounds:

![screen shot 2015-02-23 at 1 15 45 pm](https://cloud.githubusercontent.com/assets/409114/6337351/1fe340a8-bb5e-11e4-9f15-3e16aa16d5aa.png)

I made a JS Bin example illustrating the issue ~ http://jsbin.com/hasowu/5/edit?js,output

alexburner:
(if you comment out the 2 symbol lines in the jsbin, you can see the exported SVG works fine)

iconexperience:
Interesting. Here is the string representation of the svg (you can create it with `console.log(new XMLSerializer().serializeToString(svg));` ).

`<svg x="0" y="0" width="200" height="200" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><defs><symbol viewBox="-50,-50,100,100" id="symbol-1"><circle cx="0" cy="0" r="50" fill="none" stroke="#000000" stroke-width="6" stroke-linecap="butt" stroke-linejoin="miter" stroke-miterlimit="10" stroke-dasharray="" stroke-dashoffset="0" font-family="sans-serif" font-weight="normal" font-size="12" text-anchor="start" mix-blend-mode="normal"/></symbol></defs><g fill="none" stroke="none" stroke-width="1" stroke-linecap="butt" stroke-linejoin="miter" stroke-miterlimit="10" stroke-dasharray="" stroke-dashoffset="0" font-family="sans-serif" font-weight="normal" font-size="12" text-anchor="start" mix-blend-mode="normal"><use x="50" y="50" xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#symbol-1" width="100" height="100"/></g></svg>`

I found a question on stackoverflow that addresses this issue:

http://stackoverflow.com/questions/19981499/avoiding-the-clipping-of-symbols-in-svg

I tried the suggested solution and it works for your case, too. Just adding  `overflow="visible"` to the `symbol`node makes the circle draw as desired.

The question now is, how to get that attribute to the `symbol` node. You can do this directly by manipulating your svg dome structure. Or you can use the onExport callback of paper.js to add this.

[...]

PR #643, opened 2015-02-25
Fixing PlacedSymbol overflow

alexburner:
Initial value for `overflow` of svg elements is `visible`, but for elements that establish new viewports (as with these placed symbols) the value is overridden to `hidden`. A hidden overflow results in clipping of thick strokes, and may have additional consequences (I'm here because I was having trouble with my symbols' strokes).

As detailed in this issue: https://github.com/paperjs/paper.js/issues/642


lehni:
Great,thanks!

PR #644, opened 2015-02-27
Update to use jsdom v4 (document.defaultView vs. document.parentWindow).

rbvea:
jsdom recently dropped support for node.js in favor of io.js and made some [breaking changes](https://github.com/tmpvar/jsdom/blob/master/Changelog.md#400).  This fixes the latter issue when using paper.js and io.js v4+

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   },
   "dependencies": {
     "canvas": ">= 0.7.0",
-    "jsdom": ">= 0.6.0",
+    "jsdom": "3.1.2",
     "request": "~2.27.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


lehni:
Great, thanks! So if I understand this right, we should ideally be requiring different versions based on the use in io.js or node.js. But there is no way to do such a thing?

rbvea:
Yeah... you could either specify `jsdom : "3.x.x"` and have everyone stick with node or specify `latest` and have everyone migrate over to io.js.  I actually got paper.js working with io.js locally but pushing to heroku was problematic...  Maybe there should be a branch of paper.js that supports node?

lehni:
I haven't used io.js myself... Not sure how widespread it is? Most people are propaply sticking with node.js for the time being. There's also an ongoing effort to reconcile the two. I think it's worth waiting a little to see what happens: https://github.com/iojs/io.js/issues/978

PR #664, opened 2015-03-23
Make canvas an optional dependency

louisremi:
canvas should be listed as an optional dependency of paper.js as not everyone is running Paper.js inside node.js

I'm trying to setup Travis-CI for my project and build fails because of this dependency, although I don't need it. I see that I could tweak my Travis config to download a bunch of OS dependencies, but this will slow down Continuous Integration, for no added value.

Please save everyone's time with this simple PR : )

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,9 +24,11 @@
     "node": ">= 0.4.0"
   },
   "dependencies": {
-    "canvas": "~1.2.1",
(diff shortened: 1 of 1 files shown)


lehni:
Same issues here: Pull request against `master` instead of `develop`, and your email is not defined. Could you fix this request before I merge it? Thanks!

Issue #670, opened 2015-04-04
Support custom events.


lehni:
Right now, only events registered in `Emitter#_eventTypes` are supported. It would be nice if `Emitter#on()` would register new event types on the fly so they can work.

PR #674, opened 2015-04-08
custom events

rumman23:
This change enables listening to custom event types without explicit declaration in classes as proposed in #670. There are basic emitter tests as well.

### DIFF
--- a/src/core/Emitter.js
+++ b/src/core/Emitter.js
@@ -23,17 +23,15 @@ var Emitter = {
                 this.on(key, value);
             }, this);
         } else {
-            var entry = this._eventTypes[type];
-            if (entry) {
-                var handlers = this._callbacks = this._callbacks || {};
-                handlers = handlers[type] = handlers[type] || [];
-                if (handlers.indexOf(func) === -1) { // Not added yet, add now.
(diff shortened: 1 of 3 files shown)


src/core/Emitter.js
```diff
@@ -23,17 +23,15 @@ var Emitter = {
...
-                var handlers = this._callbacks = this._callbacks || {};
-                handlers = handlers[type] = handlers[type] || [];
-                if (handlers.indexOf(func) === -1) { // Not added yet, add now.
-                    handlers.push(func);
-                    // See if this is the first handler that we're attaching,
-                    // and call install if defined.
-                    if (entry.install && handlers.length == 1)
-                        entry.install.call(this, type);
-                }
+            var entry = this._eventTypes[type],
```
jardakotesovec: This assumes that _eventTypes exists, which is not true when there are no _events on class. This is case for example of Project class.

lehni:
Looks good! Will merge and then look into details, such as the comment above.

lehni:
@jardakotesovec this has been addressed in 6b428f5a37e32bff25b62261fb279f77daab542d

PR #681, opened 2015-04-18
Fix paperscript source from being executed out-of-order

hackalyze:
It was possible for loadAll to execute paperscript source in a different order from the order the script tags appear in the HTML.  This is due to the script source being retrieved asynchronously in loadScript in the call to Http.request.  This fix retrieves paperscript source synchronously unless the HTML5 async attribute is included on the script element.


lehni:
Very good! Thanks!

PR #683, opened 2015-04-21
Get canvas size from style first

jardakotesovec:
Try to read size from style first, because DomElement.getSize() returns
bounding box, which might not represent real canvas height/width when
2D/3D CSS transformation is used.

### DIFF
--- a/src/view/View.js
+++ b/src/view/View.js
@@ -58,9 +58,19 @@ var View = Base.extend(Emitter, /** @lends View# */{
         };
 
         function getCanvasSize() {
-            // Try visible size first, since that will help handling previously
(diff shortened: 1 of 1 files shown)


jardakotesovec:
I am opened to any other solution, but I had problem with growing canvas, because of skew css transformation. So this makes possibility be explicit about the size via style.

jardakotesovec:
Closing this for now, because edges of transformed canvas are not antialiased - at least in chrome, so I had to solve my use case differently anyway.  

Probably make sense to reopen this only if someone have a good use case for that, so we don't add unnecessary complexity to the code.

Issue #687, opened 2015-05-05
Gradient.clone() ignores 'radial' property


iconexperience:
If you call `clone()` on a radial `Gradient` object, the result will be a linear gradient.

Here is a <a href="http://sketch.paperjs.org/#S/jVHBasQgEP0V8ZIEgthAL5aecui10GOSg1G7kdgxqNtQlvx7NbFbKGVZvcy8mTfznl4w8A+FGX6bVRATrrGwMuUnx6VWENAzArWil5yWlx5QPj7YxTPUdcWXMsauRY0ooY9DjbpiNFzMEXgYhvqXkYZww1BwZ3WgW/V0BD0I7YRRed8rDxNpd6T81GolIi5XLg6kNFGOZvKujWmtsS7SsrIf4ewaZQHW6ZMGhjJ1sV4HbSFXpfJBA0/ItWW0Z5CeRN4U2n1/D1sP6R4dzW21hEtZ7nWr49M1lMYXqqo/Jpr7XBBhLKiy+tdNc4ed5pafHuLXj07xeUlaPWbdsH0D">simple Sketch</a> to illustrate the issue.

![image](https://cloud.githubusercontent.com/assets/2122773/7472412/b2e6cc24-f332-11e4-877d-0c2de8adf2cb.png)

I iwll try to create a pull request for the fix.

lehni:
Great, thanks! Please note that merge requests should be created to the develop branch now, not master. Will merge this one in manually now.

iconexperience:
Sorry, I still suck at using Github, will try to make it better next time. And thanks for merging.

lehni:
Not a problem!

PR #689, opened 2015-05-06
Change clone() function to fix issue #687

iconexperience:
As described in issue #687, clone() ignored the 'radial' property. Added the property to the constructor in the clone() function to get the expected behaviour.


lehni:
This was manually merged.

PR #701, opened 2015-05-29
Add support for converting Rasters to jpeg DataURL

alecdhuse:
Added support for converting Rasters to jpeg DataURL via an options
variable.

### DIFF
--- a/src/item/Raster.js
+++ b/src/item/Raster.js
@@ -447,9 +447,13 @@ var Raster = Item.extend(/** @lends Raster# */{
     /**
      * Returns a Base 64 encoded {@code data:} URL representation of the raster.
      *
+     * @param {options} options for converting to DataURL.
+     *
      * @return {String}
      */
-    toDataURL: function() {
+    toDataURL: function(options) {
+        options = options || {};
+        
         // See if the linked image is base64 encoded already, if so reuse it,
         // otherwise try using canvas.toDataURL()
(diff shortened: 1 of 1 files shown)


alecdhuse:
I did find out this can be accomplished by getting the canvas object from the raster and calling its toDataURL function with the jpeg option.

You may want to disregard this pull request.

sapics:
:+1: 

I want this feature too!

I guess that
```diff
- if (!options.quality) {
-   //default quality used by most browsers
-   options.quality = 0.92;
- }

- if ((options.type !== "image/jpeg") && (options.type !== "image/png")) {
-   //Default to image/png if options.type is an unknown type
-   options.type = "image/png";
- }

- return canvas ? canvas.toDataURL(options.type, options.quality) : null;
+ return canvas ? canvas.toDataURL(options.type, options.quality || 0.92) : null;
```
looks better for file-size and there are some possibilities that a user want to use image/webp and so on.

alecdhuse:
@sapics Your code does look much better.

sapics:
@alecdhuse Thanks!

I think that this pr makes paper.js much useful!
When authors of paperjs are agree with this pr, i guess that we need to make a pr to `paperjs/develop`.

lehni:
Yes this does look good! A PR towards the develop branch would be much appreciated indeed.

PR #702, opened 2015-05-29
SVG Import defs first

jardakotesovec:
I had problem with importing SVG from Affinity designer, because it exports defs element as last in root level. Other viewers handles that, so its apparently not against the standard. It actually just [recommend](http://www.w3.org/TR/SVG/struct.html) where it should be.

This PR just loads defs first on root level. Likely root level is sufficient. 

Let me know if you think it requires some improvement. Thanks

### DIFF
--- a/src/svg/SVGImport.js
+++ b/src/svg/SVGImport.js
@@ -87,6 +87,14 @@ new function() {
             // items and avoid calling applyAttributes() again.
             project._currentStyle = item._style.clone();
         }
+        // Put defs first, Affinity designer exports defs as last
+        if (isRoot) {
+            var defsNodes = [], otherNodes = [];
+            for (var i = 0, l = nodes.length; i < l; i++) {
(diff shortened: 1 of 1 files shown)


jardakotesovec:
Somehow one of the your commits was also included in this PR..  maybe due some rebasing. So thats fixed now..

What are your thoughts on this?

Thanks!

lehni:
This looks good! Will merge now.

lehni:
Would you mind providing a simple SVG file in the form that Affinity exports? Would like to test a bit.

jardakotesovec:
Here we go :-).
https://dl.dropboxusercontent.com/u/12126978/layers-test-affinity.svg

Thanks for merging!

lehni:
Thanks! Your fix is good but it doesn't cover all cases. defs can be anywhere really, even nested inside groups. Fixing that now too.

jardakotesovec:
Even better :-). Nice implementation! :+1: 

Thanks

PR #717, opened 2015-06-23
code minifier

sapics:
This pr is not affect to the feature.
Just a better code minifier.


lehni:
Good points, merging right away. Thanks! :)

PR #718, opened 2015-06-29
Add support for converting Rasters to dataURL (fix #701)

sapics:
This pr is just a rewrite of #701 which is created by @alecdhuse.
I guess that this would be more compatible with canvas.toDataURL.
If you think #701 is better, i would fix this pr. :)

I also add fix #664, it is not for source code, so i guess that it has little problem about CLA, isnt it?

closes #701 
closes #664

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,9 +25,11 @@
     "node": ">= 0.4.0"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


lehni:
Why did you close this?

PR #720, opened 2015-07-03
fix rectangle type name

sapics:
Shape.Rectangle type name is `rectangle`, so i guess that `rect` -> `rectangle`.


lehni:
You're right. Thanks!

PR #726, opened 2015-07-11
crossdomain image

### LINKED ISSUE #711: Safari fails on crossdomain image
My paper.js application takes a remote image:

```html
<img src="https://remote.example.com/foo.jpg" crossdomain="anonymous">
```

and applies some raster processing on it (javascript, not paperscript):

```js
var canvas = document.getElementById('myCanvas');
paper.setup(canvas);
var raster = new paper.Raster('myImg');
doStuff(raster);
```

Remote domain is adding proper CORS headers (namely, `Access-Control-Allow-Origin=*`) and indeed this works as expected on Firefox and Chrome. Safari on both OS X as well as iOS 8.3 fails on this with:

```
Unable to get image data from canvas because the canvas has been tainted by cross-origin data.
[Error] SecurityError: DOM Exception 18: An attempt was made to break through the security policy of the user agent.
```

This might or might not be a paper.js bug, but I'm curious if anyone knows why Safari isn't honoring the cross-domain policy.

sapics:
As written in #711, it would be necessary to add `image.crossOrigin = 'anonymous';`.
https://developer.mozilla.org/en/docs/Web/HTML/CORS_settings_attributes

close #711

### DIFF
--- a/src/basic/Matrix.js
+++ b/src/basic/Matrix.js
@@ -511,13 +511,12 @@ var Matrix = Base.extend(/** @lends Matrix# */{
 
     _transformCoordinates: function(src, dst, count) {
         var i = 0,
-            j = 0,
             max = 2 * count;
         while (i < max) {
-            var x = src[i++],
-                y = src[i++];
-            dst[j++] = x * this._a + y * this._b + this._tx;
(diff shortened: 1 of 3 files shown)


lehni:
This has been fixed in #754

PR #730, opened 2015-07-14
add param to raster.todataurl

sapics:
This pr is just a rewrite of #701 which is created by @alecdhuse.
I guess that this would be more compatible with canvas.toDataURL.
If you think #701 is better, i would fix this pr. :)

close #701


lehni:
I like the simplicity of this one! Merging : )

PR #749, opened 2015-08-07
Cross origin cache busting

tylor:
This is a bit of a weird one, but I wonder if others have encountered it. It also is not necessarily a Paper.js issue so feel free to disregard if it's too specific.

We're serving our JS and PaperScript from Amazon S3 on a different domain, so when our PaperScript file is loaded by XHR, it trips cross origin policies. Luckily we can enable CORS headers for our S3 bucket, which should make everything happy.

Here's where things get weird: when Safari and Chrome see the script tag with `type="text/paperscript"`, they load and cache (but don't execute) the file and its headers. Then when the PaperScript XHR loading fires, the browsers just reach into their cache and pull out the file. However, because the original script loading is not an XHR request, it doesn't include the `Origin` key in its headers and Amazon doesn't include the necessary CORS headers in their response. We get a cross origin violation, but only because the XHR is grabbing the cached file and headers.

The fix is simple but a bit of a hack: include a query string on the XHR to bust the caching.

Hope this helps!

### DIFF
--- a/src/core/PaperScript.js
+++ b/src/core/PaperScript.js
@@ -456,7 +456,7 @@ Base.exports.PaperScript = (function() {
             // project is created for it now.
             var canvasId = PaperScope.getAttribute(script, 'canvas'),
                 canvas = document.getElementById(canvasId),
-                src = script.src,
+                src = script.src + '?nocache',
                 async = PaperScope.hasAttribute(script, 'async'),
                 scopeAttribute = 'data-paper-scope';
             if (!canvas)


lehni:
I understand the issue, but don't think we can do this on all scripts by default, as it would mean that they would all get loaded twice. I need to think about an approach that solves your issue while also covering other situations well. I'll look into how coffeescript on the client side is handling this.

lehni:
I just found this on http://coffeescript.org/:

> In `<script type="text/coffeescript">` tags, to avoid possible duplicate browser requests for .coffee files, you can now use the data-src attribute instead of src.

I will implement the same solution for paperscript, as that's much better than loading the same file twice. I hope this will be acceptable for you as a solution.

tylor:
This is a great solution, thanks for looking into it!

PR #759, opened 2015-08-22
Minor fix to Matrix.js

sapics:
This is not a bug, but it would be better. :)


lehni:
Indeed! More descriptive commit messages would usually be better, but I'll merge this one in right away.

sapics:
@lehni I am sorry about commit message. I would be careful about that.

PR #760, opened 2015-08-25
Add method curve.getAnchorNearestLocation

sapics:
The method `curve.getNearestLocation` is totally awesome method!

This pr add `curve.getAnchorNearestLocation` method, and it makes `curve.getNearestLocation` precisely with less calculation, when the curve has no handles.

I make a test case at http://jsfiddle.net/sapics/3451fv3c/24/.

### DIFF
--- a/src/path/Curve.js
+++ b/src/path/Curve.js
@@ -867,6 +867,11 @@ statics: {
             minDist = Infinity,
             minT = 0;
 
+        // If there are no handles,
+        // we can get nearest location precisely with less calculation
+        if (this.isStraight())
+            return this.getAnchorNearestLocation(point);
+
         function refine(t) {
             if (t >= 0 && t <= 1) {
                 var dist = point.getDistance(Curve.getPoint(values, t), true);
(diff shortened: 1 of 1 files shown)


sapics:
Close this pr, because the structure of getNearestLocation has changed :)

lehni:
I meant to respond to this. The intention is of course right, but I'd like to merge this directly into Curve.getNearestParameter(), without the need for a separate method.

lehni:
But I think your approach is only correct if your curve handles are positioned at 1/3 and 2/3 of the line, in which case the curve-time runs linearly across the line. If the line does not define any handles, or collinear handles at different lengths (thus also appears straight), the curve-time doesn't run linearly. A faster solution would still be to project the point onto the line, and then use `getParameterOf()` for the projected point. I shall implement that.

lehni:
It may have worked for you but I believe it would have returned the wrong value for `location.parameter`, so that `location.point != location.curve.getPointAt(location.parameter, true);`

sapics:
Thanks for implementing!

I am also confirmed that my calculation (it is only work for no handles) parameter is not same as getNearestParameter.

It looks strange. So i create a test case at http://jsfiddle.net/3451fv3c/33/.
It return 0.28714072541674046.
My calculation was created that it would return 0.2.
If 0.287 is intended result, sorry, my calculation is wrong.

Sorry, my bad English.

sapics:
Sorry, i was totally wrong.
I understand what I make mistake.
I will create a new pr.

Issue #773, opened 2015-09-10
Assumptions for Self Intersections of Curves are not fully reliable


iconexperience:
When checking for self intersection of a curve, the algorithm currently follows these steps:

1. Check if the handles or negative handles of the curve intersect. It is assumed that curves can only self-intersect if this condition is fulfilled.
2. Split the curve at <code>t=0.5</code> and check if the two curve parts intersect. A curve that self intersects always forms a loop, and the point at <code>t=0.5</code> is assumed to always lie within the loop.

This works for the vast majority (maybe more than 99.9%) of all curves. But there are actually cases, where both assumptions are not true.

<a href="http://sketch.paperjs.org/#S/lZNNb6MwEIb/isUpkVjH30CqHlY5rbSHSj2GHFyYAgoxlXFSrar+9x1MqqRptdJayMB8vY9n4C1x9gDJOnncQ6jaJE2qoZ7eVysytsMrqY7+BClprat7GFcOGhu6E3wYUoJ3ElogL0PnArGBhHtGdelWq5P15MWGltwTB6/kAR8XbyM0B3BhXG+3POMpkbpICYsXL6g0JjeyYFoKrSSaOKeK5SJXTEQf36Vkq4xOiTJZSn7IXOCulJmL7Hbvy7vS/a+y1IJKVlwW1udCUK7ZZRVRWhiqZa4Mn3fUVbmk2uRMZVJmIsv5hCVprgQGM6k1E0Vh4vFEzgrkVlobPmncIMdeI/OETp87P4bNZEHvbDn2/Z9H6KEKUGNY8Ec4Zx66GqdxdVy66XzVw+KmFG0gPEyD+hkWOKU01limREwIcxGM7vvN0A8ey5XJwTbYNFsmZ6Xz3Pm11u/OwSLC0/gVcDoen4K3VThb5ySOQp+ibF3fBEwYHwp0DH7YwwWl8QAugpxDxL8gxPcQ4jOE+AIhriHEFwgPdUQoXTW4ccB+9UMztxl7+8sF8CMOqEPnwuHAlrQH14R2GXNKh3/Ykwe7j+pjst4mXCW7978=">Here is an example.</a>

![image](https://cloud.githubusercontent.com/assets/2122773/9787255/75283dca-57c1-11e5-8f22-1b3ccc11643a.png)

As you can see there is a small loop, but the handles do not intersect and the magenta colored point at <code>t=0.5</code> is far outside of the loop.

[...]

PR #793, opened 2015-09-30
Fix for #773

iconexperience:
Indroduced more reliable method for finding self intersection on curves. Moved code for finding self intersection on curve to Curve.js


lehni:
@iconexperience Great, thanks! As for the diff, I can't tell but I'll just merge it and then maybe brush things up a little. No worries at all!

PR #794, opened 2015-10-02
Fix to minimize floating point noise

sapics:
https://github.com/paperjs/paper.js/blob/develop/src/util/Numerical.js#L213 looks strange condition.
I guess that it would be right condition, isn't it?
```diff
- if (b >= MACHINE_EPSILON && b <= MACHINE_EPSILON) {
+ if (b >= -MACHINE_EPSILON && b <= MACHINE_EPSILON) {
```

EDIT: If it is convenient to make a PR to `boolean-fix`, I would create a new PR :)

### DIFF
--- a/src/util/Numerical.js
+++ b/src/util/Numerical.js
@@ -207,10 +207,9 @@ var Numerical = new function() {
             } else {
                 // No real roots if D < 0
                 if (D >= -MACHINE_EPSILON) {
-                    D = D < 0 ? 0 : D;
-                    var R = sqrt(D);
-                    // Try to minimise floating point noise.
-                    if (b >= MACHINE_EPSILON && b <= MACHINE_EPSILON) {
+                    var R = sqrt(D < 0 ? 0 : D);
+                    // Try to minimize floating point noise.
+                    if (b >= -MACHINE_EPSILON && b <= MACHINE_EPSILON) {
                         x1 = abs(a) >= abs(c) ? R / a : -c / R;
                         x2 = -x1;
                     } else {


lehni:
It sure looks suspicious, but since it involves [MACHINE_EPSILON](https://en.wikipedia.org/wiki/Machine_epsilon) and was written by @hkrish I am reluctant to change it before hearing from him about why it was written this way.

hkrish:
Yes you are right @sapics . That is what it's meant to do.

    b ≥ 0-ε && b ≤ 0+ε
    // or
    abs(b) ≤ ε

@lehni Can you please make the change? Thanks.

/Hari

lehni:
@sapics I've cherry-picked your commit into the boolean-fix branch, no need for a new PR. Will brush it up a little, as there is no need to calculate sqrt(0) when D < 0.

lehni:
@hkrish, thanks for taking care of it!

But now I have one more question. The code now looks like this:

    // No real roots if D < 0
    if (D >= -MACHINE_EPSILON) {
        var R = D < 0 ? 0 : sqrt(D);
        // Try to minimize floating point noise.
        if (abs(b) < MACHINE_EPSILON) {
            x1 = abs(a) >= abs(c) ? R / a : -c / R;
            x2 = -x1;
        } else {
            ...

But assuming that -MACHINE_EPSILON < D < 0:

- R is then 0, because:

    var R = D < 0 ? 0 : sqrt(D);

- if: (abs(b) < MACHINE_EPSILON), and: (abs(a) < abs(c)), then:

    x1 = -c / R

-> There will be a division by 0.

Is this intentional? If not, how can we circumvent it correctly?

sapics:
@lehni Thanks for better fix!

About x1 to Infinity:
As you noticed, there are some equations to diverge.
`x1 = abs(a) >= abs(c) ? R / a : -c / R;`
`x2 = c / q;`

Sorry, I don't know it is intesional or not.
But it would not be a problem because this function has `isFinite` check at the end.

hkrish:
Yup. Sorry for the mess. :)

I was trying to minimise floating point errors combining different schemes as explored by Jim Blinn in his paper published in ACM "How to solve a Quadratic Equation". Looks like I have missed a few conditions.

Since then, I was looking a bit systematically into papers devoted to the subject and as it stands, probably the best way to solve quadratics seems to be as detailed in Kahan's papers 
 1. http://www.cs.berkeley.edu/~wkahan/Math128/Cubic.pdf
 1. http://www.cs.berkeley.edu/~wkahan/Qdrtcs.pdf

So (This was a quick conversion, please do verify)

	// No real roots if D < 0
	if D >= -εMLD {
		let Q = D < 0 ? 0 : sqrt(D)
		let R = b + (b < 0 ? -1 : 1) * Q
		// Instead of comparing b to zero, we can compare b±√D to 0. ← Please remove this comment
		if R == 0 {
			x1 = c / a
			x2 = -x1
		} else {
			x1 = R / a
			x2 = c / R
		}
	}

The coefficient scaling we used so far is a crude one. To avoid floating point error in computing √(b² - ac), I would recommend switching to Kahan's *dscrmt* function (in the second paper). This needs good test schemes as well.

Probably let's keep the discriminant calculation as is. Let's continue testing he new one in parallel.

Thank you for catching the issues.

/Hari

PR #796, opened 2015-10-06
Faster simpler and more reliable implementation of getConvexHull()

iconexperience:
New implementation of getConvexHull() uses vertical distances instead of orthogonal distances, which is simpler, faster and more reliable.


iconexperience:
Sorry, needs improvement

PR #804, opened 2015-10-12
Small refactoring in getConvexHull()

iconexperience:
Make calculations of distances more concise.


lehni:
Great, thanks! Just as a little remark: I think the convention is to keep branch names short, lowercase and free from special characters, a it like css classes : )

http://stackoverflow.com/questions/273695/git-branch-naming-best-practices

PR #811, opened 2015-10-24
Faster curve.getNearestParameter when curve has no handles

sapics:
This pr make curve.getNearestParameter to faster and more precisely, when curve has no handles.
This pr is fixed version of #760.

I made a test case for check the result.
Current develop-branch build version: http://jsfiddle.net/3451fv3c/54/
This PR (294904e) build version: http://jsfiddle.net/3451fv3c/56/

### DIFF
--- a/src/path/Curve.js
+++ b/src/path/Curve.js
@@ -645,6 +645,20 @@ statics: {
     },
 
     getNearestParameter: function(v, point) {
+        if (!this.hasHandles(v)) {
+            var p1x = v[0], p1y = v[1],
+                p2x = v[6], p2y = v[7],
+                v12x = p2x - p1x, v12y = p2y - p1x,
+                t = ((point.x - p1x) * v12x + (point.y - p1y) * v12y) /
+                      (v12x * v12x + v12y * v12y);
+            if (t <= 0) return 0;
+            if (t >= 1) return 1;
+            if (!t) return 0.5;
+            var roots = [];
+            if (Numerical.solveCubic(2, -3, 0, t, roots, 0, 1) === 1)
+                return roots[0];
+        }
+
         var count = 100,
             minDist = Infinity,
             minT = 0;
--- a/src/path/Curve.js
+++ b/src/path/Curve.js
@@ -650,13 +650,11 @@ statics: {
                 p2x = v[6], p2y = v[7],
                 v12x = p2x - p1x, v12y = p2y - p1x,
                 t = ((point.x - p1x) * v12x + (point.y - p1y) * v12y) /
-                      (v12x * v12x + v12y * v12y);
-            if (t <= 0) return 0;
-            if (t >= 1) return 1;
-            if (!t) return 0.5;
-            var roots = [];
-            if (Numerical.solveCubic(2, -3, 0, t, roots, 0, 1) === 1)
-                return roots[0];
+                      (v12x * v12x + v12y * v12y),
+                epsilon = /*#=*/Numerical.EPSILON;
+            if (t < epsilon) return 0;
+            if (t + epsilon > 1) return 1;
+            return 0.5 - Math.cos(Math.acos(2 * t - 1) / 3 + Math.PI * 4 / 3) || 0.5;
         }
 
         var count = 100,


sapics:
Additional commit 5d29d86 uses solved cubic equation for faster calculation.
Test case of this: http://jsfiddle.net/3451fv3c/58/

lehni:
I can't seem to run the Fiddles. Both give me: "Uncaught ReferenceError: paper is not defined"

sapics:
Sorry, you get error at http://jsfiddle.net/3451fv3c/59/ ?

lehni:
Yeah it happens on all of those... But it works on Safari, so I can see what's going on. One question regarding 89a45a668227767375b5c054027ed68b4689526f. I remember we moved away from trigonometric solutions in `Nurmerical.solveCubic()` for reasons of imprecision. I guess in the case of a line, they are still working well? Ideally we'd work this into `Numerical.solveQuadratic()` (which is what   `Nurmerical.solveCubic()` is falling back to, and probably should have been used here), for the case of a line, right?

sapics:
>  I remember we moved away from trigonometric solutions in Nurmerical.solveCubic() for reasons of imprecision.
> I guess in the case of a line, they are still working well?
In this calculation, we need to solve only simpler equation, `2 * z^3 - 3 z^2 + t = 0`, at 0<z<1, 0<t<1.
In this case, Nurmerical.solveCubic() looks work well.

At 89a45a6, I calculate above equation with algebraically from https://en.wikipedia.org/wiki/Cubic_function.

lehni:
Oh yeah I forgot we're not dealing with a linear equation, just a curve that appears as a line : ) I need to take some time to look into this properly later.

lehni:
Could you explain the code a bit? You first project the point onto the line and calculate its linear parameter across the line, right? Could you explain where `2 * z^3 - 3 * z^2 + t = 0` is coming from then? And `z` is the curve-time parameter for the linear parameter `t` that we're looking for?

sapics:
I am sorry for less explanation.
`z` is same as `minT`.
`t` is the parameter which is I used in #760.

lehni:
Yes but could you explain the formula used for translating one to the other. Where does it come from? I'd like to add comments to the code so it's clear.

sapics:
`2 * z^3 - 3 * z^2 + t = 0` is comming from bezier cubic curve calculation.
From https://en.wikipedia.org/wiki/B%C3%A9zier_curve
```
B(z) = (1-z)^3 P_0 + 3(1-z)^2z P_1 + 3(1-z)z^2 P_2 + z^3 P_3
```
This is linear case, so B(z) is already calculated as `t` and P_0 = P_1 = 0 and P_2 = P_3 = 1 for 0 <= t <= 1 and 0 <= z <= 1
```
t = 3(1-z)z^2 + z^3
2 * z^3 - 3 * z^2 + t = 0
```

lehni:
Perfect, thanks! Here's a thought: What if this approach was moved to `getParameterOf(point)`, and `getNearestParameter(point)` would just project the point onto the line in case of `isStraight()`, and then call  `getParameterOf(point)` with this projected point? The advantage is a speed-up of `getParameterOf(point)` for linear curves, and I don't think `getNearestParameter(point)` would be much slower then. Thoughts?

lehni:
I am going to merge this in and then add some comments, but we should keep discussing this.

lehni:
One more question: Can you explain the ` || 0.5`?

sapics:
Thank you for trying to read my bad English!

As you noticed, if we use `getParameterOf(point)`, we can move `(!hasHandles()) -> isStraight()` which would be more wide condition.
I used algebraically solution is just for speed. If `getParameterOf(point)` has enough speed, I am totally agree with using `getParameterOf(point)`. It would be better!

`|| 0.5` is for `t` is `NaN` :)

sapics:
I guess that most of the case is `t < 0` or `t > 1`.
Thus, there might be little performance difference by using `getParameterOf(point)`.

lehni:
No problem for the english, I understand it, and really appreciate your input!

So I think you get `NaN` when `Math.acos()` produces `NaN`, meaning the passed value's range (`2 * u - 1`) is outside [-1, 1]. Shouldn't we catch this before doing the calculations?

lehni:
Hmm and hat's the case when `t` is outside [0, 1], which we already handle, so you should never get `NaN`?

sapics:
I guess that when `(v12x * v12x + v12y * v12y) = 0` and `((point.x - p1x) * v12x + (point.y - p1y) * v12y)=0`, then, it would be `NaN`, isn't it?

sapics:
`NaN < 1` and `NaN > 0` is always `false` :)

lehni:
Yes that's right. So we should handle the case of P1 = P2 and just return 0

sapics:
The calculation `0.5 - Math.cos(Math.acos(2 * NaN - 1) / 3 + Math.PI * 4 / 3)` is `NaN`, so catch this before or later might be ok.

sapics:
BTW, is it better `(!hasHandles()) -> isStraight()` by using `getParameterOf(point)`?

lehni:
Yes, if we move this special case for `!hasHandles()` to `getParameterOf(point)`, then we catch more linear cases. Will work on this.

sapics:
I am totally agree!

I guess that we needs a few fixes as,
```javascript
        if (!this.isStraight(v)) {
            var p1x = v[0], p1y = v[1],
                p2x = v[6], p2y = v[7],
                v12x = p2x - p1x, v12y = p2y - p1x,
                t = ((point.x - p1x) * v12x + (point.y - p1y) * v12y) /
                      (v12x * v12x + v12y * v12y),
                epsilon = /*#=*/Numerical.EPSILON;
            if (t < epsilon) return 0;
            if (t + epsilon > 1) return 1;
            return Curve.getParameterOf(v,
                     new Point((1 - t) * p1x + t * p2x, (1 - t) * p1y + t * p2y));
        }
```

lehni:
Before we tackle that, there are some issues with the current code.  While your example works as it should (here simplified):

```js
var p1 = new Point(100,100);
var p2 = new Point(200,100);
var noHandleCurve = new Curve(p1, new Point(), new Point(), p2);

var test = function(num){
    var x = p1.x + num;
    var pt = noHandleCurve.getNearestPoint(x, p1.y);
    console.log(x, pt.x, pt.x - x);
}

for(var i = -1; i <= 101; ++i)test(i);
```

This produces wrong results:


```js
var p1 = new Point(100,110);
var p2 = new Point(200,110);
var noHandleCurve = new Curve(p1, new Point(), new Point(), p2);

var test = function(num){
    var x = p1.x + num;
    var pt = noHandleCurve.getNearestPoint(x, p1.y);
    console.log(x, pt.x, pt.x - x);
}

for(var i = -1; i <= 101; ++i)test(i);
```

The only difference being the y-coordinate of the line.

sapics:
Thank you for fixing more readable!
Oh sorry, i made a mistake at `v12y = p2y - p1x` => `v12y = p2y - p1y` , which you have fixed current develop branch.

lehni:
haha oh dear yeah didn't catch that : )

sapics:
Oh, not yet, sorry.
`vy = p2y - p1x;` => `vy = p2y - p1y;`

lehni:
So I've played around a bit with moving this code to `Curve.getParameterOf()`, but it starts breaking boolean operations again. @hkrish explained before that using the trigonometric method can lead to quite large imprecisions, which was the reason for the current implementation of `Numerical.solveCubic()` that he wrote. The trigonometric method is probably precise enough for `getNearestParameter()` which otherwise is just using an iterative approach. So we can either keep it as it is, or we can handle all straight curves and use the slower  `Curve.getParameterOf()`

lehni:
Although, I shouldn't rush to conclusions. I should look into the cases where the results are different, triggering the issues. Other things could be going on...

lehni:
I think your calculation of the point for `getParameterOf()` here is wrong: https://github.com/paperjs/paper.js/pull/811#issuecomment-150778666

This yields great results: `Curve.getParameterOf(v, new Point(p1x + u * vx, p1y + u * vy));`

sapics:
Ups, sorry many mistakes, and i need to go out now.
`Curve.getParameterOf(v, new Point(p1x + u * vx, p1y + u * vy));` looks work well!

PR #812, opened 2015-10-24
Remove unnecessary calculation

sapics:
I guess that `if` condition is deletable, isn't it?


lehni:
Haha, nice catch. Yeah it sure looks like it is! But I'd like to ask @hkrish first to find out why it was there to begin with. Perhaps the comparison was meant to be performed with MACHINE_EPSILON tolerance?

lehni:
@hkrish, could you have a look at this one please? Thanks!

hkrish:
Good catch! That is unnecessary.

lehni:
Sweet : )

PR #814, opened 2015-10-27
Handle exportSVG({ precision: 0 }) correctly

johan:
At present, `path.exportSVG({ precision: 0 })` actually picks precision 5 – have it only do that when the parameter is omitted / `undefined` / `null`.


lehni:
Thanks!

johan:
And same! I wasn't sure what steps to run to rebuild the appropriate parts of the codebase and actually run the test suite, but I hoped you might have those set up with Travis or similar to vet the pull request – or, if not, run them prior to next release.

lehni:
Not using Travis yet but it's easy enough the run the unit test directly on the develop branch, see README. All tests pass : )

Issue #835, opened 2015-11-25
Simple subtract operation throws open path error message


iconexperience:
While working on path offsetting I found that very simple subtract operations on `CompoundPath` objects result in an open path error message:

```javascript
// compound path
var cp = new CompoundPath();
var p1 = new Path.Rectangle(100, 100, 100, 100);
cp.addChild(p1);
cp.strokeColor = "green";
// path
var p2 = new Path.Circle(180, 180, 50);
p2.strokeColor = "red";
// subtract
var c = cp.subtract(p2);
```

The message is *Boolean operation resulted in open path segments = 2 length = 65.8395631458437*

The problem does not occur if p2 is subtracted directly from p1 (`var c = p1.subtract(p2)`).

Note that I cannot test against the `develop` branch at the moment, but the bug occurs in 0.9.25. If this is fixed in the new branch already please just close this issue.

iconexperience:
This is really strange. At the beginning of `computeBoolean()` we reduce both paths, so `cp` is replaced by a simple `Path`, which should be identical to `p1`. But obviously something is different.

iconexperience:
I found out that when we use the `CompoundPath`as `path1` in `computeBoolean()`, it's reduced clone `_path1` unexpectedly has a parent (which is the 'CompoundPath`). This happens, because `reduce()` is called after `clone()`, and `reduce()` does not remove the resulting path if the original compound path has no parent.

Maybe in `CompoundPath.reduce()`the line

```path.insertAbove(this); ```

should be changed to

```this.parent ? path.insertAbove(this) : path.remove(); ```

iconexperience:
As an alternative I thought about changing the behaviour of `insertAbove()`, but I think the current behaviour is correct: If there is no parent, `insertAbove()` should fail and null should be returned.

So here is my summary:
- If `reduce()` is called on an item, the resulting item should have the same parent as the original item. This implies that if the original item has no parent, the result should have no parent, too.
- Currently if the original item has no parent, the resulting item can have a parent, because `insertAbove()` does nothing if there is no parent, so the child that will replace the original item will keep its parent.
- Since the behaviour of `insertAbove()` is correct, the `reduce()` function has to make sure that the parent is set to null if the original item has no parent.

`CompoundPath.reduce()` does not need to be changed, because if the `CompoundPath` is replaced by an empty `Path`, the new `Path`object initially has no parent.

Only `Item.reduce()` needs to be changed. I will create a pull request.

[...]

PR #846, opened 2015-12-04
Fix for item.reduce()

iconexperience:
Fix for #835


lehni:
There was something strange with this pull request. I've attempted a manual merge, but I think you got lost as the author along the way... 3072cef2a4e60c391a28fbcf82059ff646f507c4

iconexperience:
No problem, this was a tiny change and the only important thing is that the issue is fixed.

PR #848, opened 2015-12-07
getWinding()-fix for #819

### LINKED ISSUE #819: Wrong hit result
I got the hit result `{ type: 'fill', item: Path@1 } ` where definitely no hit result should be.

wrong hittest

http://jsfiddle.net/x5t1v10e/

iconexperience:
Handle cases when point y is equal to y of horizontal curve in path. Fixes #819

### DIFF
--- a/src/path/PathItem.Boolean.js
+++ b/src/path/PathItem.Boolean.js
@@ -326,75 +326,121 @@ PathItem.inject(new function() {
         } else {
             var xBefore = px - epsilon,
                 xAfter = px + epsilon;
-            // Find the winding number for right side of the curve, inclusive of
-            // the curve itself, while tracing along its +-x direction.
-            var startCounted = false,
-                prevCurve,
-                prevT;
-            for (var i = 0, l = curves.length; i < l; i++) {
-                var curve = curves[i],
-                    values = curve.values,
-                    winding = curve.winding;
-                // Since the curves are monotone in y direction, we can just
-                // compare the endpoints of the curve to determine if the
-                // ray from query point along +-x direction will intersect
-                // the monotone curve. Results in quite significant speedup.
-                if (winding && (winding === 1
+            var loopStartIndex,
+                loopEndIndex = 0;
+            do {
+                // Determine beginning and end of loop and the first and last curve with
+                // non-zero winding within the loop
+                loopStartIndex = loopEndIndex; // index of first curve in loop
+                loopEndIndex = loopStartIndex + 1; // index after last curve in loop
+                var curve,
+                    firstWindCrv = null, // first curve in loop with winding != 0
+                    lastWindCrv; // last curve in loop with winding != 0
+                for (var i = loopStartIndex, l = curves.length; i < l; i++) {
+                    curve = curves[i];
+                    if (curve.winding) {
+                        if (!firstWindCrv)
+                            firstWindCrv = curve;
+                        lastWindCrv = curve;
+                    }
+                    if (curve.next !== curves[i + 1]) {
+                        loopEndIndex = i + 1;
+                        break;
+                    }
+                }
+                // walk through single loop of curves
+                var startCounted = false,
+                    prevWindCurve, // non-horizontal curve before current curve
+                    nextWindCurve, // non-horizontal curve after current curve
+                    prevT = null;
+                curve = null;
+                for (var curveIndex = loopStartIndex; curveIndex < loopEndIndex; curveIndex++) {
+                    if (!curve) {
+                        prevWindCurve = lastWindCrv;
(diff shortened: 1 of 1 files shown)


iconexperience:
This changes was a bit complicated, so I decided to rework part of `getWinding()` completely. Here are some remarks regarding the changes:

- in the new code we first detect the start and end of each loop and then run through each loop separately. This way we do not have to check if we are on the same loop all the time.
- horizonal curves (these are absolutely horizontal curves) are usually ignored. Only if the point is on the curve and the winding before the curve is different from and after the curve we treat this as a "touch point".
- Since we ignore horizontal curves now, an intersection at the end of one curve, which is again found at the beginning of a following curve, is now counted only once, even if any number of horizontal curves are between these two curves.

iconexperience:
What I do not fully understand (and what I did not dare to touch) are the cases when `horizontal == true`. Can you tell me when this is used?

Another thing that I do not fully understand is the calculation of the tangents at the previous and next curve to find touch points (lines 371 to 375 in the original code). Since all curve are monotone in y direction, wouldn't it be possible to simply use the winding of the curve instead of calculating the tangent?

iconexperience:
And here is a little more background on the problem that this patch is supposed to fix. If the point happens to have the same y value as the end of one curve, the same intersection is found twice because the start of the next curve has the same y value. This was handled previously by checking if the intersections found at the beginning of a curve is the same as the one at the end of a curve. So cases like these were not a problem:
image
But the old code failed if there were horizontal curves (absolutely horizontal lines) between two curves. In this case the curves sharing the intersection were not adjacent. So in cases like this, the two intersections were counted:
image
There can be cases when several horizontal curves are between two intersection points, but things get really funny if the first and last curve of a path are horizontal, like in this case:
image
The new code should solve all thes cases. I will try to provide some more tests.

iconexperience:
I have created a test for a path consisting of only horizontal curves. <a href="http://sketch.paperjs.org/#S/lVNNj4IwEP0rTfcARIKgq5vgbvbgHzC7hz0ABz5GaOy2plQ9GP/7ThFFZA/apEDfvJm+eS1HKtJfoCH93oDOK+rSXBZmvU8V2aa6WkkmdE0+SBSrWBAcUfDquyTw/cTtIH8ITR5hTR5KnA5ZsxYxQLIwTzMb1ShWwIGsUL3dteA0pK23ZpwvJZcKaTF9AYCYtqFaK7mBm2CWZXfBH1boCoNBi+Zc1lAgoNUOLkzQP0wUTJRfOw52TGEPQhZFTJ2r0IvYiul/HJ5Msbt537p7A+ZD46ZD6G3oZQ/qrBuPSaHSg1FEto0kg67RCtsIZSjQX+DrvdPscRClrhAdjRxyNPzTpdyziUaboedS6JSJxlO097yq7eZATa59LRKxJMIuSA8IEudssqmH2bXk4HFZ2oyM8ERDnPhx3aTjNnszlXO4uT7eskGe2d0lsxsBTXrvynX9fRKrVADCIiGxFBTW4mogaQf+jpmCdHM+DxpGyekP>Here is the sketch</a>

iconexperience:
<a href="http://sketch.paperjs.org/#S/lVNdb4IwFP0rTfcgREJA5zZ1yx78A2Y++AA88HGFRmxNqS6L8b/vtoogsoeRlLbn3tN7eg+cKI93QGd0tQWVFtShqcj0/hhLso9VsRSMq4p8kCCUISf4BP7Ic8iLFzkN8tyL+F4bevsDabO8XqQ+OZrrtx5GHYri8E2WqNJqpNomae9uWFkuRCkkpoX0CQBCeg1VSoottIJJknSCa5apAoP+FU1LUUGGgJIHqDNBrRnPGM+/DiVYIYUjcJFlIbXvhaK2lYqlagl2F0ymSNI6ZaVWkO+AK3evr+CQ8fUSNbFzmVzCz1VvXaNgqsepMTZv1G7ntAv4ky4y7lo5ffD2IcV/7XVJx+r1BrVbWilDhd4cp/dGtFsCz1WB6HBok1PN1emp4Cpm3PQebbjsKsv0UXOt2yEBiwKUQO4AP7IvvdTnIbsSJbilyC1GhtjJGQ5c3Io0uaa28ajHtX9Ud8ikJcDQ79xs7vdJBmgs8AGZkYGEbGBo524z9Yy/aSIh3prPpaKzIDr/Ag==">Here is another test case</a> where the path's first and last curve are horizontal. Seems to work well with the new code.

iconexperience:
@lehni 
This seems to be still buggy. Please do not merge, I will try to fix it and issue a new pull request.

PR #852, opened 2015-12-14
getWinding-fix(2)

iconexperience:
getWinding()-fix for #819


iconexperience:
Compared to #848 this handles the cases correctly when the curves array has length zero. One case when this can occur is if a path has been scaled to zero before a boolean operation.

lehni:
Thanks a lot for this, @iconexperience! I've merged this in now, to do further tests.

PR #856, opened 2015-12-17
Fix shape.strokeBounds with strokeScaling is false

### LINKED ISSUE #697: mouse collision detection doesn't take strokeScaling into account
I noticed this when using onMouseOver to change the colour of objects. The raycast used for the mouse events seems to always assume strokeScaling is set to true.

Simple repro butchered from the hit testing example here:

http://goo.gl/kTaAzW

If you turn stroke scaling on and off you can see the difference.

sapics:
This pr fix shape.strokeBounds with shape.strokeScaling=false.

### DIFF
--- a/src/item/Shape.js
+++ b/src/item/Shape.js
@@ -270,10 +270,16 @@ var Shape = Item.extend(/** @lends Shape# */{
     },
 
     _getBounds: function(getter, matrix) {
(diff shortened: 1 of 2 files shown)


sapics:
Travis work!

lehni:
Looks reasonable. I am going to merge it in and then test a bit further. Thanks!

Issue #859, opened 2015-12-17
Intersection not found on crossing curves


iconexperience:
Although the boolean code looks fairly stable now, there are still cases that do not work. In this simple example `path.subtract()` yields in a wrong result.

```javascript
var p1 = new Path({segments:[
    [230, 360, 0, 0, 0, 0],
    [326.04999999999995, 361.95, 0,0, 7.100000000000023, 5.300000000000011],
    [347.74999999999994, 377.3, -7.300000000000011, -5, 0,0],
    [260, 400],
], closed:true});
var p2 = new Path({segments:[
   [329.8373529833907, 360.99927475751736, 0, 0, -0.5084518552435497, 1.7136677994218417],
    [327.9816245617005, 366.0401044369074, 0.7293957729680756, -1.645975116174725, 0, 0],
    [300, 380]
], closed:true});
var r = p1.subtract(p2);
p1.fillColor = 'rgba(255, 0, 0, .2)';
p2.fillColor = 'rgba(0, 0, 255, .2)';
r.fullySelected = true;
```

<a href="http://sketch.paperjs.org/#S/fVLbTuMwEP2VKC8UKR15bI8nLuKJH1iJx9KHNDVQkSaVky4CxL/vuKE0C4hRFCvHZy7nTN7yttqFfJHfPoWhfsyLvO426ftvFbM9ZtdZG56zP9XwOHvrw8MutEO/WN7FuzaTWGqjisw4eZ2fVfF5bbQDZf05KLER0qkKITOgmoQ2RUZgphDitJ5l4Gk9K/WYQdLm/DVPsLHNpIBOo1r1Aa2KrG66PmwWQzyE98urBB6F69+FizAPpWFD2pfGeMVHF0Am0myZmJCNOzkyV0CqtIQlkbaGrBc6AqNxjllEaCwt8n++MfgSnbbkkJU62pasVKisNU4ainIFrL3xxHK4UjFJxzmCs+RlAJRMy5q+L0WlnZVq9asFURzYI/SH9RCrepjt9Xgn2P22aW66pkuUi/iwrmaa6CQW9OXFSNQ/EEfOkf5JjHB/aJqX29CEeggb4aZRxkG24Rleu24noDkjtSwjpKIR1t2h3fQfyNVJ4xjyM69jqJ723VaWly+Wq/d/">Here is the sketch</a>

Here is how it looks (the two paths at the left, the actual sketch with the result selected at the right):

[...]

PR #861, opened 2015-12-18
Improve-addCurveIntersections()

iconexperience:
In addCurveIntersections() handle cases where tDiff==0, some smaller changes and improvements

### DIFF
--- a/src/path/Curve.js
+++ b/src/path/Curve.js
@@ -1427,7 +1427,7 @@ new function() { // Scope for intersection using bezier fat-line clipping
     }
 
     function addCurveIntersections(v1, v2, c1, c2, locations, param,
-            tMin, tMax, uMin, uMax, oldTDiff, reverse, recursion) {
+                                   tMin, tMax, uMin, uMax, oldTDiff, reverse, recursion) {
         // Avoid deeper recursion.
         // NOTE: @iconexperience determined that more than 20 recursions are
         // needed sometimes, depending on the tDiff threshold values further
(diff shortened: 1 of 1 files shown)


iconexperience:
I also changed the condition for subdiving the curve to the condition recommended in the original Sederberg/Nishita algorithm (tDiff > 0.8). This makes the oldTDiff argument unnecessary, but I left it in there because I did not want to change the internal api befor e we have tested the new condition with the old edge cases..

iconexperience:
I had to replace `tDiff > Numerical.EPSILON` with `tDiff > 0` to make the example in #859 work. I have not found any negative side effects on our big test case, but I admit the whole algorithm is still a bit shaky.

lehni:
Thanks! I will have a close look tomorrow.

Looking quickly through the diff view, I realized something that I've seen before, and should maybe point out: There are quite a few changes that are just reformatting of existing code and don't relate to the actual fix. It would be better to refrain from such reformatting for a couple of reasons: The code currently follows a convention that's in place across the library and has its reasons to be the way it is, but mainly these changes make it harder to find out what actually changed functionally. Do you know what I mean?

iconexperience:
Yes, maybe I should have split this into two. But changing the order of the conditions (which is the biggest change) has a functional reason, which is avoiding unnecessary curve splitting if an intersection may already be found. This is in lines 1475 to 1490, which has only been moved up.

The changes for the fix are actually the last `else if`and `else` statements and the introduction of `v1New`, which was necessary to have access to the original `v1` at the end.

Other changes are:
- changing `t` to `u`. This is only cosmetic and makes understanding the code a bit easier.
- simplifying all those calculations of average parameter values (for example `t = tMinNew + (tMaxNew - tMinNew) / 2` to `t = (tMinNew + tMaxNew) / 2`)
- changing the condition for curve splitting to `tDiff > 0.8`. This could have been done later.

And all those changes to comment indents were not intentional. I will try to take better care on that.

iconexperience:
@lehni I will replace this pull request with two others, so we have the bugfix separated from the other code improvements.

iconexperience:
Closing this as #863 and #864 have been created as replacements. Changing the condition for subdividing the curve from `tDiff > 0.5 && oldTDiff > 0.5` to  `tDiff > 0.8` is not covered by the replacements. This can be done later once we have more tests for edge cases.

PR #883, opened 2016-01-02
Improve fat line clipping to fix #882

### LINKED ISSUE #882: addCurveIntersections() very slow for quasi-straight curves
It is known that fat-line clipping has problems with straight, collinear curves. Since we handle straight curves without fat-line clipping, this is usually not a problem for us, but there are certain cases that cause trouble.

Look at this harmless looking pair of curves:

```javascript
var crv1 = new Curve([87.34388521634617, 0, 76.75904513292927, 0, 73.54763183962362, 0, 75.20314002403848, 0]);
var crv2 = new Curve([0, 0, 0, 0, 162.51045673076925, 0, 160.90144230769232, 0]);
console.log("ints : " + crv1.getIntersections(crv2).length);
```

If you <a href="http://sketch.paperjs.org/#S/VZC7bsMwDEV/RdBko4YgkXo6yJSpc8c4g+sKThBXKmTVHYr+e+XYQFGQCw94eUl+09C/e9rSl7vPw5U2dIhva730iQxpEeRIgv8ip8+0+OpsDUOJ1ioQGqUWpiG8IUYzoxyXSiC4EjtFpqTRKCw6Dahho4oBRyE5B8nRSlvopT50qQu7Jfy35A/ZlkIDU6IYaYPcaAdqx5w5LqSEjSL8DR1imOPk2RTHqqO3kGfSko6Sp8d1bPT5OWSfZj/kW2mt1gVqNvkw5mt9KP94Tb6/f8RVSdvz5ecX">run the code</a>, you will notice that finding the intersection takes very long (a few seconds on my machine).

The reason is that the fat lines of the curves are actually thin lines (width = 0), and they are perfectly collinear. So each curve lies within the fat line of the other curve. Subdividing the curves does not help, because this results in two subcurves, that are again collinear to the original curve. So each of these curves' fat line is again within the other curve's fat line. As a result we build an enormous tree of subcurves with **16,777,215** calls to `addCurveIntersections()`.

The reason why the curves are not treated as straight lines is that while the handles and end points are perfectly aligned, one (or both) handle(s) is outside of the curve's end points (these curves can occur when scaling a path to zero in one direction). I think it is correct to not treat these curves as straight, therefore we need to find a solution within `addCurveIntersections()`.

As it turns out, the solution is very simple. We already calculate the distances for all points and handles to the line that runs through the end points of curve 2. If all these distances are zero, the curves are perfectly collinear and we can stop our search.

I will create a pull request for the fix of this issue.

iconexperience:
Add a break condition in fat line clipping to prevent very slow clipping for two curves that have all points and controlpoints exactly aligned.

### DIFF
--- a/src/path/Curve.js
+++ b/src/path/Curve.js
@@ -1462,6 +1462,9 @@ new function() { // Scope for intersection using bezier fat-line clipping
             bottom = hull[1],
             tMinClip,
             tMaxClip;
+        // Stop iteration if all points and control points are exactly collinear.
+        if (d1 == 0 && d2 == 0 && dp0 == 0 && dp1 == 0 && dp2 == 0 && dp3 == 0)
+            return;
         // Clip the convex-hull with dMin and dMax, taking into account that
         // there will be no intersections if one of the tvalues are null.
         if ((tMinClip = clipConvexHull(top, bottom, dMin, dMax)) == null ||


lehni:
Looks good! BTW: The paper.js convention is === for strict numeric comparisons, so I'll change it to that after.

iconexperience:
Thanks. But I just started reading the Git book, because I still suck at using Git.

lehni:
I should read that too : )

lehni:
(BTW: Really lovin' these automatic Travis unit test checks!)

PR #895, opened 2016-01-08
Further simplify and explain winding calculation in getWinding() (2)

iconexperience:
Further simplify and explain winding calculation in getWinding()

### DIFF
--- a/src/path/PathItem.Boolean.js
+++ b/src/path/PathItem.Boolean.js
@@ -353,11 +353,11 @@ PathItem.inject(new function() {
                 // Retrieve and use it here (See _getMonoCurve()).
                 var curve = curves[start],
                     last = curve.last,
-                    // Get the values of to the end x coordinate and winding of
-                    // the last non-horizontal curve, which will be the previous
-                    // non-horizontal curve for the first curve of the loop.
+                // Get the values of to the end x coordinate and winding of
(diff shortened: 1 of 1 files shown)


lehni:
Merging and tweaking things a bit right now.

lehni:
I've left some comments in d186c2d35626c8a9a8922d656bd5c1767d9c2419 regarding coding guidelines.

lehni:
Here's another code compression we could do... Not sure how readable it would be though (I don't struggle with these, but many people do):

```js
var x = py === yStart
    ? values[0]
    : py === yEnd
    ? values[6]
    : Curve.solveCubic(values, 1, py, roots, 0, 1) === 1
    ? Curve.getPoint(values, roots[0]).x
    : null;
```

lehni:
Or like this, if it helps reading (I think it does):

```js
var x = py === yStart ? values[0]
    :   py === yEnd   ? values[6]
    : Curve.solveCubic(values, 1, py, roots, 0, 1) === 1
        ? Curve.getPoint(values, roots[0]).x
        : null;
```

Ternary Operators FTW!

iconexperience:
Yep, that is nice (I prefer the second snipplet). If people want to understand how winding calculation works, they should know ternary  operators :)

lehni:
:+1: :)

PR #900, opened 2016-01-11
Update Curve.js, Path.js and Item.js

unpete:
fix getNearestPoint error on uninserted curves
fix insertAbove / insertBelow error on uninserted items


lehni:
Thanks! Due to changes in my code I had manually merged these in now, but your commits should still appear correctly.

PR #907, opened 2016-01-16
Improve performance of _getMonoCurves() by filtering more curves

iconexperience:
Curves are always monotone in y direction if the y coordinates of their end points and control points are sorted (yP1 >= yC1 >= yC2 >= yP2 or yP1 <= yC1 <= yC2 <= yP2). Performing this check is very cheap compared to the actual finding of extrema in y direction.

In my test case about 75% of the yet unfiltered curves can be filtered.

### DIFF
--- a/src/path/PathItem.Boolean.js
+++ b/src/path/PathItem.Boolean.js
@@ -924,8 +924,10 @@ Path.inject(/** @lends Path# */{
                 y1 = v[3],
                 y2 = v[5],
                 y3 = v[7];
-            if (Curve.isStraight(v)) {
-                // Handling straight curves is easy.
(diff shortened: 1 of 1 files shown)


lehni:
Very cool! Will merge right away : )

lehni:
Unrelated, but perhaps useful. Did you see this?

https://guides.github.com/introduction/flow/

iconexperience:
Thanks, will read. Was there anything unusual with my pull request (except for the boolean op on the wrong line :) )?

lehni:
No, not at all! That's what I meant with "not implying anything" : )

iconexperience:
Good. I feel I am getting better, but there is still a long way...

PR #910, opened 2016-01-17
Add periods to match other comments

jshen212:
- Added periods to match other comments


lehni:
While I of course appreciate you taking the time to look through the code and create a PR, I wonder if this particular PR was really worth your time? Wouldn't it be better spent trying to fix one of the many issues in the database? There are many other places where sentences were left without a period, and I don't really care. Sometimes I change that when I fix other things right next to it. But if we start creating and merging pull requests for such things, where does it end : )

PR #911, opened 2016-01-18
Add stroke hit-test option for shape items

sapics:
This is just a personal proposal.
It looks natural for me to make hitTest options.stroke activate for `shape` items.

It might be better to add `options.shapeStroke` whose default is true, to avoid making differences of result of hitTest.


lehni:
You are of course right about this. Thanks! `options.shapeStroke` is not necessary.

sapics:
Thank you for your approval!

PR #915, opened 2016-01-19
Fix matrix invert calculation

sapics:
This pr would fix matrix.invert calculation.


lehni:
You're right! That must have sneaked in during the recent refactoring. Thanks!

PR #926, opened 2016-01-26
Minor minification

sapics:


lehni:
fc2c90485087508b364abdea0f23b6aacdbc62cc makes a lot of sense, but I'm not sure I like 095cd47c1c57fe16662621a844e89216b5d8311d. It doesn't reduce the amount of additions, and barely reduces the amount of code, but becomes less readable. But I have an idea for a simpler version that I like, will accept the request and then change it.

sapics:
:+1: https://github.com/paperjs/paper.js/commit/a38834b81dee5c344fe1b2b7a2d8a9c722339cad

PR #930, opened 2016-01-29
skip arc subpaths of svg paths that start and end at same point

ludwick:
See: https://www.w3.org/TR/SVG/implnote.html#ArcImplementationNotes

Basically, if an arc in an SVG path ends where it started, it should be ignored, but it was not being which resulted in an attempt to read `getDirectedAngle` on a null [here](https://github.com/paperjs/paper.js/blob/ed8a904b74c5875b24fd8c083923f33552c5da2d/src/path/Path.js#L2573).

FWIW, I don't actually know the SVG spec (or paper for that matter) very well so I could be off base here. That said, we're seeing SVGs exported from tools such as Illustrator that have these pointless sub-paths.


sapics:
@ludwick
I am afraid that `(destPoint - current) > 1e-12` would always false, because `destPoint` and `current` are not Number but Point.
I guess that `(destPoint.subtract(current)).getLength() > 1e-12` is what you want, isn't it?

lehni:
Yes, PaperScript isn't active in core...

ludwick:
Thanks all! I'll have a look when I get into work. The unit test DID fail without the change but maybe it's erroneous. I'll confirm, fix and update.

ludwick:
Yep. Definitely is a `NaN`  which is always `<= 1e=12`. I added an additional arc directive that should be correctly processed and it's skipped. No other tests failed but I guess that could just coverage. I'll fix how this is working and make sure the test case incorporates an arc that should be included and one that should be skipped.

ludwick:
I actually need to abandon this for now due to other priorities. Sorry for the churn.

lehni:
No problem, I can look into it, it should be fairly easy.

lehni:
I'm going to reopen this, just to keep it on the radar.

sapics:
@ludwick Thank you for testing!
No problem. Thanks to you, I could find a bug in #931 :)

PR #938, opened 2016-02-02
Simplify getWindings() by using only the 'last' property to understand when a new loop starts.

iconexperience:
Simplify getWindings(), The first curve of a loop always has the 'last' property set. We use this to know when a loop starts. The very rare cases when a loop only consists of horizontal curves are not a problem, because these curves are ignored. Using this simplification also get's rid of the 'length' property, which was set in _getMonoCurves().


lehni:
Great, looks good!

PR #944, opened 2016-02-05
Change the implementation of getWinding() again to pass all unit tests.

iconexperience:
This new version does not count any intersections if the point is on a y-monotonic curve, but at the final calculation it sets the winding to at least 1 if the point is on a curve.

### DIFF
--- a/src/path/PathItem.Boolean.js
+++ b/src/path/PathItem.Boolean.js
@@ -314,10 +314,10 @@ PathItem.inject(new function() {
             py = point.y,
             windLeft = 0,
             windRight = 0,
-            isOnCurve = false,
             length = curves.length,
             roots = [],
-            abs = Math.abs;
+            abs = Math.abs,
+            isOnCurve = false;
         // Horizontal curves may return wrong results, since the curves are
         // monotonic in y direction and this is an indeterminate state.
         if (horizontal) {
@@ -357,12 +357,13 @@ PathItem.inject(new function() {
                     values = curve.values,
                     yStart = values[1],
                     yEnd = values[7];
-                // The first curve of a loop holds the last curve with non-zero
-                // winding. Retrieve and use it here (See _getMonoCurve()).
+                // The first curve of a loop holds the last curve with
+                // non-zero winding.
+                // Retrieve and use it here (See _getMonoCurve()).
                 if (curve.last) {
                     // Get the end x coordinate and winding of the last
                     // non-horizontal curve, which will be the previous
-                    // non-horizontal curve for first curve in the loop.
+                    // non-horizontal curve for the first curve in the loop.
                     prevWinding = curve.last.winding;
                     prevXEnd = curve.last.values[6];
                 }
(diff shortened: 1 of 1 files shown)


iconexperience:
This was expected to fail because the unit test `Path.contains() (straight curves with zero-winding)` counts some points on a curve as being outside the path.

@lehni: Maybe I should have combined the two pull requests. Combined they should have passed the Travis checks.

iconexperience:
Here is another test that fails with the current code, but not after the new change:

image

```javascript
var c1 = new Path.Circle({center: [200, 200], radius: 100});
var c2 = new Path.Circle({center: [150, 200], radius: 50});
var cp = new CompoundPath({children: [c1, c2], strokeColor: "blue"});
var point = new Point(100, 200);
new Path.Circle({center: point, radius: 3,  fillColor: "red"});
console.log("contains point: " + cp.contains(point));
```

<a href="http://sketch.paperjs.org/#S/fZDBbsMgEER/ZcXJUS3LpMrFVU/+gUg9hhwIbGtksmsBbg5R/j3YTtJWVXtBMNo3M+xZkD6iaMRbj8l0ohSG7fT+1AGMhFcgPMFWp65qXTAei7NBShga2K3ruoR87EsI2roxNiDr+rJ6UUHRzK//5eXmF7/5gQ83vOXjwCPZySbznfM2IGUHI8sckvmYAvfYsudsrMTBj6jEN6uBHaV7meleyFv5ZebPkjP41e+5BHh33j+SAtpHkGGK7LHy/FGovEdK2lFcLKZZeMpfqu56MeurBc17PwTU/axF0ez2lys=">Here is the Sketch</a>

lehni:
So I have a question:

On [line 396](https://github.com/paperjs/paper.js/blob/develop/src/path/PathItem.Boolean.js#L396), you test if the point is on the horizontal curve like this, comparing against zero with `< 0`. I don't fully understand this check, could you explain it?
```js
(px - x) * (px - prevXEnd) < 0)
```
On [line 409](https://github.com/paperjs/paper.js/blob/develop/src/path/PathItem.Boolean.js#L409), you have the same check, but with `<= 0`. intentional?
```js
(px - values[0]) * (px - values[6]) <= 0
```

lehni:
And regarding https://github.com/paperjs/paper.js/pull/944#issuecomment-180370569, I shall turn this case into a unit test! We have to protect ourselves against regressions in edge cases like these.

lehni:
@iconexperience and yes, it would have been better to just create one PR. a PR can contain many commits. In fact, even after opening a PR here, all the new commits to your branch that you created the PR for will appear here, until I merge the PR in. And since the two topics are so closely related, one PR would have been more suitable.

Another topic: It's good if the commit messages are quite short. It's fine to write more information into a commit message, but it's better to keep the first line short, and then add a 2nd paragraph with the details, like this:
```
Change the implementation of getWinding() again, so we pass all tests

This new version does not count any intersections if the point is on a y-monotonic curve, but at the final calculation it sets the winding to at least 1 if the point is on a curve.
```
This is also to prevent GitHub force-breaking the first line. But it's just considered good form, generally.

iconexperience:
@lehni If the distances `px - x` and `px - prevXEnd` are have opposite signs, it means that `x` is between `px` and `prevXEnd`. By using `<0`instead of `<=0` we exclude the cases `px == x` and `prevXEnd == x`, but these are already handled in <a href="https://github.com/paperjs/paper.js/blob/develop/src/path/PathItem.Boolean.js#L383">line 383</a>, so we save one character :)

image


But in line 409 we need to include the end points of the line, otherwise we would miss this point:

image

iconexperience:
And thanks for all the help on becoming a better Github citizen :)

lehni:
: )

PR #970, opened 2016-02-14
Improve getWinding() to fix #968

### LINKED ISSUE #968: Subtract operation results in open path
It looks like https://github.com/paperjs/paper.js/commit/3348fb736f142b0df42bf4d9238347b63bf98e07 has introduced a regression. The following rather simple case now throws a *Boolean operation resulted in open path* message.

``` javascript
var p1 = new paper.Path({segments:[
    [352, 280, 0, -26.5, 0, 0],
    [352, 356, 0, 0, 0, 2.1999999999999886],
    [348, 360, 2.1999999999999886, 0, -72, 0]
    ], closed:true});
var p2 = new paper.Path({segments:[
   [352, 344, 0, 0, 0, 0],
   [352, 356, 0, 0, 0, 2.1999999999999886], 
   [348, 360, 2.1999999999999886, 0, 0, 0], 
   [232, 360, 0, 0, -2.1999999999999886, 0]
   ], closed:true});

p1.strokeColor = "green";
p2.strokeColor = "red";

var res = p1.subtract(p2);

res.fullySelected = true;
```

The result actually looks fine, but still the error message indicated that something is wrong.

iconexperience:
Improve handling of points on the path in getWinding(). Fix for #968

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "node": ">=0.10.0 <6.0.0"
   },
   "dependencies": {
-    "jsdom": "git://github.com/lehni/jsdom.git#bd86dd16666241edd5f25bc4f4e144f24da3c889",
     "source-map-support": "^0.4.0"
   },
   "optionalDependencies": {
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "node": ">=0.10.0 <6.0.0"
   },
   "dependencies": {
+    "jsdom": "git://github.com/lehni/jsdom.git#bd86dd16666241edd5f25bc4f4e144f24da3c889",
     "source-map-support": "^0.4.0"
   },
   "optionalDependencies": {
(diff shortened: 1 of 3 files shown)


iconexperience:
Please take a look at `leftInstersVal`, which contains `x` and `winding` for the leftmost intersection. I do not like the name (but could not find a better one), and I am not sure if it's alright for you to use an array for storing the values like this.

lehni:
Looks great! Is there a reason you closes this again? I'm out for a while it will have a close look later. In case you're still struggling with git: have a look at SmartGit. It's a great cross platform GUI that has helped me a lot getting into git. I'm still using it regularly. And it's free for non commercial use.

iconexperience:
It still contained an error (I could have changed it), but it also included my change to packages.json an some undos for the svg package. My Git clumsyness is getting really embarrassing.

lehni:
Don't worry about it : ) But this really helps me: http://www.syntevo.com/smartgit/

iconexperience:
I will take a look. But I have it all integrated in my WebStorm, which is great. I just do the wrong things sometimes...

PR #971, opened 2016-02-14
Fix for #968 - Improve handling of points on paths in getWinding()

### LINKED ISSUE #968: Subtract operation results in open path

iconexperience:
Improve handling of points on paths in getWinding() to fix #968.
Windings for points on the path are now calculated separately and only applied if overall winding is zero.

### DIFF
--- a/src/path/PathItem.Boolean.js
+++ b/src/path/PathItem.Boolean.js
@@ -427,14 +427,14 @@ PathItem.inject(new function() {
                 }
             }
             // Use the on-curve windings if no other intersections were
-            // found or if they canceled each other.
+            // found or if they canceled each other. On single paths
+            // this ensures that the overally winding is 1 if the
+            // point was on a monotonic curve.
             if (windLeft === 0 && windRight === 0) {
                 windLeft = windLeftOnCurve;
                 windRight = windRightOnCurve;
             }
         }
-        // If the point was on a monotonic curve, we are on the path by
-        // definition. In this case ensure that the winding is at least 1.
         return Math.max(abs(windLeft), abs(windRight));
     }
 
--- a/src/path/PathItem.Boolean.js
+++ b/src/path/PathItem.Boolean.js
@@ -305,7 +305,6 @@ PathItem.inject(new function() {
             py = point.y,
             windLeft = 0,
             windRight = 0,
-            isOnCurve = false,
             length = curves.length,
             roots = [],
             abs = Math.abs;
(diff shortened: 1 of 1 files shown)


lehni:
I just realized something is odd about `leftIntersVal`: You never seem to set this to anything? Is it even still required? I have a feeling we can remove it completely!

lehni:
@iconexperience this still works: 8d5c922c50743bd8a2b5d7e00541e1bb4b40ff32

iconexperience:
Oh dear, everything worked even after I removed that part.

What do you mean by "this still works"?

iconexperience:
OK, I understand. I will explain in a minute.

iconexperience:
I think this only works because we do not have enough test cases. Anyway, the winding number is wrong although it may not even make a difference in many cases. See this example:

image

In the current implementation the red dot will have a winding of 0 to the left (no intersection) and right (windings cancel each other). Therefore we will use the `windLeftOnCurve` and `windRightOnCurve` values. But they are now `+2` and `-2`, because we do not distinguish between left and right winding paths. The should actually both be zero.

lehni:
But my change and your code don't do anything different, you agree?

lehni:
It would be good to find a case where it fails still and build a test for it, then.

iconexperience:
Yes, it was a mistake on my side. I somehow deleted that part, tested, everything worked, and I forgot to put it in again. I will try to find a test case, intersect operations could do it.

I may need until tomorrow, but the current version seems to work fine.

lehni:
Yes it works well! There is no rush with it. I've squashed tons of bugs BTW, very close to 0.10.0, thinking of rolling it out in the next 2 days or so.

iconexperience:
Yes, I was watching. That was quite impressive and I am not aware of any open issues that affect me.

There may be one last issue regarding the winding of paths in the results of boolean operations. I hope I can do this tomorrow, too.

lehni:
@iconexperience you'll probably be happy about this one: 21033f7850965cf185cd2753ba275bc6de88dec7

PR #976, opened 2016-02-15
Fix #974 - Keep empty clip mask in SVG export

### LINKED ISSUE #974: Empty clip mask should not be removed in SVG export
If a `Group` has an empty path as the clip mask, the other child items are not visible. This is the expected behaviour, because the visible area is zero.

But if the `Group` is exported as SVG, the empty clip mask will be removed and no clipping will be applied at all in the SVG, so all previously hidden child nodes become visible.

Here is a little example:

```javascript
var g = new Group();
g.addChild(new Path());
g.addChild(new Path.Circle({center: [100, 100], radius: 50, fillColor: "red"}));
g.clipped = true;

console.log(new XMLSerializer().serializeToString(project.exportSVG()));
```

<a href="http://sketch.paperjs.org/#S/dU9BasMwEPyK0EkGI9JDLw45+ZBLCgWXUohzUKSNvc1WK9ZyWxLy9yghPfYyMDPLzOxZR/cFutHdEbIfda09hxv/dqIGtVIRftRaeE6mWvZxsC6EdkQK5ma8ujya6h/DtiiewJw9xAzSqO3TYlGrArtaiQs4T416LsoBiVomLie9Fgi9vjwyPWFKEMqMLDMUqY+e48QElni4N328bDoQdIQnEFPZ6Y+8cZcF42CS8Cf4bOE3seTufV0WV/es8u1ewB0TY8yTbra7yxU=">And here is the Sketch</a>

If you open the SVG output in a browser, you will see that the red circle is displayed. The SVG output is here:
```SVG
<svg x="0" y="0" width="1705.00012" height="1530.00012" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <g fill="none" fill-rule="nonzero" stroke="none" stroke-width="1" stroke-linecap="butt" stroke-linejoin="miter" stroke-miterlimit="10" stroke-dasharray="" stroke-dashoffset="0" font-family="none" font-weight="none" font-size="none" text-anchor="none" style="mix-blend-mode: normal">
        <g>
            <path d="M50,100c0,-27.61424 22.38576,-50 50,-50c27.61424,0 50,22.38576 50,50c0,27.61424 -22.38576,50 -50,50c-27.61424,0 -50,-22.38576 -50,-50z" fill="#ff0000"/>
        </g>
    </g>
</svg>
```

As you can see, the clip mask is gone.

In detail this happens because in <a href="https://github.com/paperjs/paper.js/blob/develop/src/svg/SvgExport.js#L72">SvgExport.js line 72</a> the function `exportSVG()` returns `null`. 

To fix this a check if the path is a clip mask could be added in <a href="https://github.com/paperjs/paper.js/blob/develop/src/svg/SvgExport.js#L113">SvgExport.js line 113</a> like this:

```
if (segments.length === 0 && !item.isClipMask())
```

iconexperience:
A path should be exported in SVG if it serves as a clip mask, otherwise all items in the clip group that were hidden by clipping in paper.js become visible in the SVG. This fixes issue #974.

### DIFF
--- a/src/svg/SvgExport.js
+++ b/src/svg/SvgExport.js
@@ -110,7 +110,7 @@ new function() {
         var segments = item._segments,
             type,
             attrs = getTransform(item._matrix);
-        if (segments.length === 0)
+        if (segments.length === 0 && !item.isClipMask())
             return null;
         if (matchShapes && !item.hasHandles()) {
             if (segments.length >= 3) {


lehni:
Hmm. One could also argue that empty parts should be exported in any situation. I think that's actually better!

Issue #975, opened 2016-02-15
paper-core with nodejs errors out.


georeith:
Get the following error when importing `paper-core.js` in node.

```
/paper/dist/node/extend.js:21
    paper.PaperScript.sourceMapSupport = {
                                       ^

TypeError: Cannot set property 'sourceMapSupport' of undefined
    at module.exports (/paper/dist/node/extend.js:21:40)
```

PR #979, opened 2016-02-17
paper.PaperScript is undefined in paper-core.js.

louisremi:
I got `Cannot set property 'sourceMapSupport' of undefined` errors when testing plumin.js in Node.js


lehni:
Yes that has already been fixed: #975 

I wasn't testing paper-core.js, just paper-full.js. could you try again with a latest prebuilt version?

louisremi:
Ok, my bad! I did pull before writing the fix, but didn't realize it was already there.

lehni:
No worries at all! Other than that, it works for you? Glad I finally found some time to wrap things up. v0.10.0 is imminent and should be a very good release!

PR #984, opened 2016-02-18
Fix namespaces of exported svg

aschmi:
Hi there,

this fix address an issue I've had with the `exportSVG` function in Internet Explorer 11 (on Windows 7). The resulting SVG appeared to be corrupted. The problem was that the svg-tag had invalid namespace attributes:

The following result of exportSVG

`<svg` **`xmlns="http://www.w3.org/2000/svg"`** `x="0" y="0" width="870" height="870" version="1.1"` **`xmlns:NS1="" NS1:xmlns:xlink="http://www.w3.org/1999/xlink“`**`>…</svg>`

should have been 

`<svg x="0" y="0" width="870" height="870" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink“>…</svg>`

I found a description of the exact same problem as well as a solution in this [SO](http://stackoverflow.com/questions/19610089/unwanted-namespaces-on-svg-markup-when-using-xmlserializer-in-javascript-with-ie)

### DIFF
--- a/src/svg/SvgElement.js
+++ b/src/svg/SvgElement.js
@@ -23,7 +23,9 @@ var SvgElement = new function() {
         // Mapping of attribute names to required namespaces:
         attributeNamespace = {
             href: xlink,
-            xlink: xmlns
+            xlink: xmlns,
+            xmlns: xmlns,
+            'xmlns:xlink': xmlns
         };
 
     function create(tag, attributes, formatter) {
--- a/src/svg/SvgElement.js
+++ b/src/svg/SvgElement.js
@@ -18,8 +18,8 @@
 var SvgElement = new function() {
     // SVG related namespaces
     var svg = 'http://www.w3.org/2000/svg',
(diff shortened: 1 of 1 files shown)


lehni:
Great, thanks for spotting! For future reference, it would be good to create an issue for the problem, and then reference it from the commits that fix it, so the changes are explained. The current commit messages are not very clear, as they don't mention IE at all. But tis' fine, I'll add code comments after merging.

lehni:
The added slashes break SVG importing in some cases (see for example the failed Travis CI build: https://travis-ci.org/paperjs/paper.js/builds/110370095). Could you explain why the slashes are necessary?

lehni:
I found a fix for it. IE appears to be very inconsistent. The slash seems to be only required when calling `setAttributeNS()`. When adding it for `getAttributeNS()`, it fails.

lehni:
This fix breaks the `examples/Scripts/BlendModes.html` example. Opening again...

PR #1024, opened 2016-03-24
nextSeg can ben null in findBestIntersection

louisremi:
I haven't been able to create a reduced test case yet, but on numerous occasions I have seen findBestIntersection crash on L512 because `seg.getNext()` returned null on L511.
Checking the truthiness of nextSeg before accessing its properties fixed my problems without causing any tests to fail.

### DIFF
--- a/src/path/PathItem.Boolean.js
+++ b/src/path/PathItem.Boolean.js
@@ -509,7 +509,7 @@ PathItem.inject(new function() {
             while (inter) {
                 var seg = inter._segment,
                     nextSeg = seg.getNext(),
-                    nextInter = nextSeg._intersection;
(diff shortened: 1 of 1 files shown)


lehni:
This would only happen if the paths in question are open. A test case of paths where this happens would be appreciated!

louisremi:
I'll get back to this next week with a reduced test case if I can.

lehni:
Please do : )

lehni:
@louisremi would be great still if you could find a test-case for this!

louisremi:
We fixed some code on our side and I wasn't able to reproduce the problem after that. So I guess it wasn't exactly a bug in Paper after all. I'll re-open if I can reproduce it.

PR #1037, opened 2016-04-15
exportSVG: avoid TypeError: Converting circular structure to JSON

unpete:
`item._data` may contain data that can not be serialized to JSON
To avoid TypeError: Converting circular structure to JSON(…), added option `excludeData`


lehni:
I don't think I am going to merge this in. If you don't want data to be serialized, just don't store in in the data property, but somewhere else.

PR #1067, opened 2016-06-09
Add support for absolute coordinates in getPathData()

iconexperience:
**Update: This does not work, see my last comment**

This adds support for absolute coordinates in getPathData() in Path with only minimal changes to the code and full backwards compatibility. An extra parameter `_absolute` is added to indicate if abolute coordinates should be used.

To use this for SVG export the `_absolute` parameter would have to be added to the Compound Path and the options would need to support this, too.

The reason why absolute coordinate are desired sometimes is that due to imprecisions it can happen with relative data that the start and end point on a closed path are not identical. This cannot happen with absolute coordinates.

### DIFF
--- a/src/path/Path.js
+++ b/src/path/Path.js
@@ -294,7 +294,7 @@ var Path = PathItem.extend(/** @lends Path# */{
     // Enforce bean creation for getPathData(), as it has hidden parameters.
     beans: true,
 
-    getPathData: function(_matrix, _precision) {
+    getPathData: function(_matrix, _precision, _absolute) {
         // NOTE: #setPathData() is defined in PathItem.
         var segments = this._segments,
             length = segments.length,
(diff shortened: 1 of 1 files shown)


lehni:
Hmm. the thing is, getPathData() is not supposed to be called directly, really. The documented way is `Path.pathData`as a property. This is just the getter for it... Maybe we should just switch to absolute all the time?

iconexperience:
Using only absolute data would make the code even a little bit smaller :)

I really do not like the relative path coordinates. Converting a path that is available in absolute coordinates to relative and then after loading again back to absolute just does not feel right. And it has the problem described above.

Relative coordinates may reduce the length of the sting a little bit because values are often smaller, but with the floating point numbers this should not make much difference. 

In my opinion switching to absolute coordinates would be the correct thing to do. I will create a pull request and you can decide if you want to use it.

iconexperience:
Closed this because it does not work correctly. It only creates curves even for lines. All tests were passed because the tests only check if the result path is correct.

PR #1068, opened 2016-06-09
Use absolute coordinates in Path.getPathData()

iconexperience:
Change Path.getPathData() so it uses absolute coordinates instead of relative ones. This solves the problem with closed paths using relative coordinates that sometimes the last point is different from the start point, which then results in an extra point when reimporting the data.

### DIFF
--- a/src/path/Path.js
+++ b/src/path/Path.js
@@ -321,14 +321,14 @@ var Path = PathItem.extend(/** @lends Path# */{
                 // appropriate and resulting in shorter strings.
                 if (inX === curX && inY === curY
                         && outX === prevX && outY === prevY) {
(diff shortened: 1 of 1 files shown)


lehni:
I have made some tests, using the infamous SVG tiger, using relative and absolute SVG coordinates. The results:

* Relative: 76'688 bytes.
* Absolute: 98'348 bytes.

I don't think the savings of 22% are all that small... We should think again if this is really the way to go. It may also break backward compatibility. There are other ways of fixing the issue of `closePath()`.

I do agree that we probably want support for absolute coordinates, but it should be hidden behind an `options` switch and not made default (yet).

I will look into this, no need for more PRs for now.

PR #1077, opened 2016-06-14
Fix variable leak

sapics:
Fix variable leak in src/node/extend.js.
The other fixes are just corrections in indentation.


lehni:
Tiptop!

sapics:
Gern geschehen!

Issue #1078, opened 2016-06-15
Chrome cannot optimize Numerical.solveQuadratic()


iconexperience:
This is not an important issue at all, but I think it is quite interesting and I have a solution ready that even results in slightly shorter code.

While profiling some Paper.js operations with Chrome I noticed a little warning icon next to the <a href="https://github.com/paperjs/paper.js/blob/develop/src/util/Numerical.js#L217">`Numerical.solveQuadratic()`</a> function's data.

![image](https://cloud.githubusercontent.com/assets/2122773/16079733/593fa290-3305-11e6-8f39-0f0fac05c11b.png)

The lillte icon indicates that Chrome's V8 failed to optimize the function.

Hovering over the icon gave the extra information *Not optimized: Optimized too many times*. This again means that Chrome's V8 tried to optimize it several times, but the behavior was inconsistent, so it gave up eventually.

After some googling I found that wrong handling of `NaN`s can lead to failed optimizations and further digging showed that this part of the function was the culprit:
```javascript
var mult = pow(10,
    abs(Math.floor(Math.log(gmC) * Math.LOG10E)));
if (!isFinite(mult))
    mult = 0;
``` 
As I understand this (and tests indicate that I am correct)  `isFinite(mult)` can only be false if `gmC <= 0`. In this case, already the `Math.floor()` is called with a `NaN` or `Infinity` argument. Actually, the way `gmC` is calculated it cannot be negative (but it can be zero), so I tried changing the part of the code to
```javascript
var mult = gmC === 0 ? 0 : pow(10,
    abs(Math.floor(Math.log(gmC) * Math.LOG10E)));
``` 

This passes all tests and results in V8 being able to optimize `Numerical.solveQuadratic` function!

lehni:
Great detective work, @iconexperience. And very interesting to read about this V8 stuff. I spent quite a bit of time a while back optimizing for it, but haven't looked back into all this. Thanks!

iconexperience:
Just for the record and for future reference, there are some other functions that currently cannot be optimized, such as `Point.isClose()`, `Segment.setHandleIn()`, `Rectangle.touches()` and others. The reason why these functions cannot be optimized is that the `arguments` are 'leaked' to another function. More information on why this blocks optimization and on workarounds can be found at  https://github.com/petkaantonov/bluebird/wiki/Optimization-killers

[...]

PR #1079, opened 2016-06-15
Make Numerical.solveQuadratic() optimizable for Chrome V8

iconexperience:
Little change to help Chrome's V( optimize `Numerical.solveQuadratic()`. In my use case this resulted in an overall performance improvement of about 1.5%. For details see #1078

### DIFF
--- a/src/util/Numerical.js
+++ b/src/util/Numerical.js
@@ -234,10 +234,8 @@ var Numerical = new function() {
                     // We multiply with a factor to normalize the coefficients.
                     // The factor is just the nearest exponent of 10, big enough
                     // to raise all the coefficients to nearly [-1, +1] range.
-                    var mult = pow(10,
-                            abs(Math.floor(Math.log(gmC) * Math.LOG10E)));
-                    if (!isFinite(mult))
-                        mult = 0;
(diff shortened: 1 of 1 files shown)


src/util/Numerical.js
```diff
@@ -234,10 +234,8 @@ var Numerical = new function() {
                     // We multiply with a factor to normalize the coefficients.
                     // The factor is just the nearest exponent of 10, big enough
                     // to raise all the coefficients to nearly [-1, +1] range.
-                    var mult = pow(10,
-                            abs(Math.floor(Math.log(gmC) * Math.LOG10E)));
-                    if (!isFinite(mult))
-                        mult = 0;
+                    var mult = gmC === 0 ? 0 : pow(10,
+                        abs(Math.floor(Math.log(gmC) * Math.LOG10E)));
```
lehni: I tiny detail: We've been using double-indents to distinguish a broken statement from a nested code block (see left). This used to be a Java convnetion that found its way into paper.js and is in use across the code-base...

iconexperience:
Makes sense, and I know this from working with Java a while back.

PR #1081, opened 2016-06-16
Fix to cover the case when Numerical.solveQuadratic return -1

sapics:
Just a very very minor case fix in path/PathItem.Boolean.js and Numerical.solveCubic :)


lehni:
Good catch. お疲れ様です!

sapics:
Danke schön!

PR #1082, opened 2016-06-16
Add rough bounds checking for _addBounds()

iconexperience:
Performance of `_addBounds()` can be improved significantly by performing a rough bounds checking first. This is a cheap way to check if the curve can extend the current min or max values at all. Only if the current bounds can be extended by the curve's bounds, further calculation needs to be done.
Also, if the values of a curve are sorted, the extrema are simply the start and end point.

### DIFF
--- a/src/path/Curve.js
+++ b/src/path/Curve.js
@@ -785,42 +785,54 @@ statics: /** @lends Curve */{
      */
     _addBounds: function(v0, v1, v2, v3, coord, padding, min, max, roots) {
         padding /= 2; // strokePadding is in width, not radius
-        // Code ported and further optimised from:
-        // http://blog.hackers-cafe.net/2009/06/how-to-calculate-bezier-curves-bounding.html
-        function add(value, padding) {
-            var left = value - padding,
-                right = value + padding;
-            if (left < min[coord])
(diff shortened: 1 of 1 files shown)


lehni:
Looks good. A suggestion though:

Instead of writing a long description into the commit message (those should stay short and concise) it's better to add it as a code comment to the actual block of code.

And when indenting code, please note that we don't let code go over 80 chars. If you're using Sublime Text, there's a great plugin to treat comments automatically, down to the formatting of @param & co: https://github.com/ehuss/Sublime-Wrap-Plus

lehni:
This one broke jshinting, because `function add(value, padding)` is inside a block now. Will fix it shortly:

`W082  Line 801, Col 13  Function declarations should not be placed in blocks. Use a function expression or move the statement to the top of the outer function.`

lehni:
But hey, really great optimization!

iconexperience:
Yep, this has quite an effect. Regarding the function I actually wasn't sure where to place it correctly, so I am not too surprised something in the tool chain complained.

lehni:
Inner functions are supposed to be outside of any scopes, not exactly sure why, but that's what the linters want, so we comply :)

PR #1083, opened 2016-06-16
Add check of value range in Curve.solveCubic to improve performance

iconexperience:
Performance of `Curve.solveCubic()` can be improved by first checking if the specified value is within the curve's value range. If it is outside the range, the expensive call to `Numerical.solveCubic()` is not necessary.


lehni:
Renaming your branches in my commit messages now :)

Issue #1085, opened 2016-06-17
Inconsistent calculation in Numerical.solveCubic and Numerical.solveQuadratic


sapics:
In [this sketch](http://sketch.paperjs.org/#S/hZBNC4JAEIb/yrInC1NX6CJ0ig5BWOFRPYzbUKLuxvhxif57a5J26INhYD6e4X2ZG1dQIQ94VGAjL9zmUp/6vgNiyFYsbCukXELpbA7RdrcPbUbCzOPUFP6zSNQE1brscN1mubQ8ZzlHmy3eksSMuS6TUKP4fmVYMSX5442fKKmVgdEp9dmiYZuoPnq/f40dWzgRNEbGmPFGTz84MXAvpY/y5mkZIRRXnaum5kGc3h8=), you can confirm that solutions of two equivalent equation are not same in Numerical.solveCubic and Numerical.solveQuadratic.
It is just a special case problem, thus, it would be enough to fix after 0.10 :)

## Numerical.solveCubic

The solution of `a x^3 + b x^2 + c x + d = 0` is calculated in Numerical.solveCubic.
https://github.com/paperjs/paper.js/blob/develop/src/util/Numerical.js#L302

When the case1 (`a = 0.5*EPSILON, b = c = d = - EPSILON`) then, (because of `abs(a) < EPSILON`)
it is converted to `b x^2 + c x + d = 0` in https://github.com/paperjs/paper.js/blob/develop/src/util/Numerical.js#L307.

On the other hand, the solution with case1 is equivalent to the solution with `a = 0.5, b = c = d = - 1` (case2) where Numerical.solveCubic return right solution.
In the case2, it is NOT converted to `b x^2 + c x + d = 0`, becase of `abs(a) > EPSILON`.

The conversion in case1 causes problem.
I guess that we need to fix the conditions of this conversion in (https://github.com/paperjs/paper.js/blob/develop/src/util/Numerical.js#L307 and https://github.com/paperjs/paper.js/blob/develop/src/util/Numerical.js#L312).

## Numerical.solveQuadratic

Similar problem is occured in Numerical.solveQuadratic.
`a x^2 + b x + c = 0` with `a = EPSILON, b = 0, c = -EPSILON` which solutions are `-1` and `1`, but Numerical.solveQuadratic return no solutions.
On the other hand, Numerical.solveQuadratic could return right solutions for equivalent equation `a = 1, b = 0, c = -1`.

## One of the solutions (but it might cause some other problem...)

[...]

PR #1087, opened 2016-06-20
Improvements to solve polynomials in Numerical.js

sapics:
This PR would make better calculation to Numerical.solveCubic and Numerical.solveQuadratic.
This fix is just implemented @hkrish [c code](https://gist.github.com/hkrish/9e0de1f121971ee0fbab281f5c986de9) in js code.
For details, please see #1085.

### DIFF
--- a/src/util/Numerical.js
+++ b/src/util/Numerical.js
@@ -238,6 +238,7 @@ var Numerical = new function() {
                         abs(Math.floor(Math.log(gmC) * Math.LOG10E)));
                     a *= mult;
                     b *= mult;
+                    B *= mult;
                     c *= mult;
                     // Recalculate the discriminant
                     D = b * b - a * c;
--- a/src/util/Numerical.js
+++ b/src/util/Numerical.js
@@ -356,8 +356,8 @@ var Numerical = new function() {
             }
             // The cubic has been deflated to a quadratic.
             var count = Numerical.solveQuadratic(a, b1, c2, roots, min, max);
-            if (isFinite(x) && count >= 0
-                    && (count === 0 || x !== roots[count - 1])
+            if (isFinite(x) && (count === 0
+                    || count > 0 && x !== roots[0] && x !== roots[1])
                     && (min == null || x > min - EPSILON && x < max + EPSILON))
                 roots[count++] = min == null ? x : clamp(x, min, max);
             return count;
--- a/src/util/Numerical.js
+++ b/src/util/Numerical.js
@@ -255,7 +255,7 @@ var Numerical = new function() {
             if (D !== 0 && abs(D) < MACHINE_EPSILON) {
                 // If the geometric mean of the coefficients is small enough
                 var sc = (abs(a) + abs(b) + abs(c)) || MACHINE_EPSILON;
-                sc = pow(2, -Math.floor(Math.log(sc) / Math.log(2) + 0.5));
+                sc = pow(2, -Math.floor(Math.log(sc) * Math.LOG2E + 0.5));
                 a *= sc;
                 b *= sc;
                 c *= sc;
@@ -333,7 +333,7 @@ var Numerical = new function() {
                 // Scale the coefficients by a multiple of the exponent of
                 // coefficients so that all the bits in the mantissa are
                 // preserved.
-                var p = pow(2, - Math.floor(Math.log(s) / Math.log(2)));
+                var p = pow(2, -Math.floor(Math.log(s) * Math.LOG2E));
                 a *= p;
                 b *= p;
                 c *= p;
--- a/src/util/Numerical.js
+++ b/src/util/Numerical.js
@@ -341,12 +341,12 @@ var Numerical = new function() {
             }
             // If a or d is zero, we only need to solve a quadratic, so we set
             // the coefficients appropriately.
-            if (a === 0) {
+            if (abs(a) < EPSILON) {
                 a = b;
                 b1 = c;
                 c2 = d;
                 x = Infinity;
-            } else if (d === 0) {
+            } else if (abs(d) < EPSILON) {
                 b1 = b;
                 c2 = c;
                 x = 0;
--- a/src/util/Numerical.js
+++ b/src/util/Numerical.js
@@ -67,6 +67,25 @@ var Numerical = new function() {
         return value < min ? min : value > max ? max : value;
     }
 
+    function splitDouble(X) {
+        var bigX = X * 134217729, // X*(2^27 + 1)
+            Y = X - bigX,
+            Xh = Y + bigX; // Don't optimize Y away!
+        return [Xh, X - Xh];
+    }
+
+    function higherPrecisionDiscriminant(a, b, c) {
+        var ad = splitDouble(a),
+            bd = splitDouble(b),
+            cd = splitDouble(c),
+            p = b * b,
+            dp = (bd[0] * bd[0] - p + 2 * bd[0] * bd[1]) + bd[1] * bd[1],
+            q = a * c,
+            dq = (ad[0] * cd[0] - q + ad[0] * cd[1] + ad[1] * cd[0])
+                    + ad[1] * cd[1];
+        return (p - q) + (dp - dq);
+    }
+
     return /** @lends Numerical */{
         TOLERANCE: 1e-6,
         /**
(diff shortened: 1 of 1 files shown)


src/util/Numerical.js
```diff
@@ -255,7 +255,7 @@ var Numerical = new function() {
             if (D !== 0 && abs(D) < MACHINE_EPSILON) {
                 // If the geometric mean of the coefficients is small enough
                 var sc = (abs(a) + abs(b) + abs(c)) || MACHINE_EPSILON;
-                sc = pow(2, -Math.floor(Math.log(sc) / Math.log(2) + 0.5));
+                sc = pow(2, -Math.floor(Math.log(sc) * Math.LOG2E + 0.5));
```
hkrish: You can replace `Math.log(sc) * Math.LOG2E`  with `Math.log2(sc)`

sapics: Thank you for your comment!
Unfortunately, IE looks not supported.
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Math/log2.

sapics: The result would be changed a little,
`sc = Math.exp(-Math.floor(Math.log(sc) + 0.5));`
is better?

lehni:
Thanks @sapics! Great work!

sapics:
My pleasure! Thanks!

lehni:
Unfortunately the RR has broken a unit test in the node test suite. I haven't had time to look at it yet, but you can run it yourself: `gulp test:node`. It's weird because the same test doesn't fail in the browser...

sapics:
From the [sketch](http://sketch.paperjs.org/#S/lVNNa9tAEP0ri6DYJvYyHzv7kdBTzwVDe7N8UBSFCDur4Mgtbch/70h2aqd1DxVC7L55szPvzeqlyNVjU1wXXzZNXz8U86Lu7ob9t2pnlmg+mtx8N8uqf5iuymz0WZHA3CDr5/Su52/B4CwxJ+EUPbuASk3OIjiXIIL4kJB+pykuPgZAcSElSjHMzQIsCAMq02NMIpTAn51PFlH5QwyR0Q0FvI0YhUm0KJITxSwH8CHqqd4Ri9IWZAU5+eQoeGZyxz7WZV7Pbso8Kqb/VxzJagmPTjgMDfuDZJDIGF0cVHF4y1t4i8H7hEmQCFWm0hdo1R/EROIEdcX8T0NBLjgqo5cYvWdPEgSCQ89HL5OQ5xDJR08x8pnoMi/R3rfb7adu2+1U+eR2u28m6sWS3uO75m6E0XZPVd32PxRUzmkDVjR+NJHPTTzoKIvPF5T8LWQ7tAwqhSCONyJF0jHPR5iRfAIBFySpQFcriHps0vmyRgh8OlJlIOukFQjqM5qByhyUmUBpFJ3Of6RqkRjRQQC9RBDN5QYu1f9ZFmUers6S/7SxqjeTNz/awZ8y3+9z3bddNv109nJwRfNOFk6vrtqZ+WDo5hB8bvqv7WPT7ftpPzcCAEOp1zJrvi70V73dNdXmqWtz/1xcr9avvwA=), the result in node looks right?

default

lehni:
I may be an issue with the unit tests in Node.js, perhaps even a rendering issue with node-canvas. I'll investigate.

lehni:
Now I get it. You're right, the expected result in the test was wrong. Funny how the same test in the browser didn't fail. Rasterizing imprecisions...

sapics:
Yes, it's really strange. :confused:
Both would use same source file paper-full.js...

Anyway,  in this test-case, your cleanups in https://github.com/paperjs/paper.js/commit/02658c9e743685973b6779d60116adc28008df04 and https://github.com/paperjs/paper.js/commit/7e3d18f5d4149e0b25c3675050ed9682be9a4499 gives us same result in node and browser! :+1:

lehni:
@sapics those changes should be purely aesthetic, and reduce the amount of code a bit by reusing code. The test passes because of 1914e64e4bfa6e2952d450705e0579007389f65f

sapics:
Yes, test:node passes by https://github.com/paperjs/paper.js/commit/1914e64e4bfa6e2952d450705e0579007389f65f.

But now, browser test/index.html also returns `M274.23395,194.10449l-0.00118,-0.00313c0.01275,-0.00503 0.01337,-0.00299 0.00118,0.00313z` :)

lehni:
Oh, interesting! I thought that was already the case before my changes? Hmm, it was rather late last night : )

sapics:
It's really interesting!

Probably, it would be after your changes. :)

Here, there was not `+ 0.5`, but now it has `+ 0.5`.
https://github.com/paperjs/paper.js/commit/02658c9e743685973b6779d60116adc28008df04#diff-2791193c45cac795fddd8b95ddb71503L336
It might make change?? :)

sapics:
`s < 1e-7` -> `s < 1e-8` might changes the result :)
https://github.com/paperjs/paper.js/commit/02658c9e743685973b6779d60116adc28008df04#diff-2791193c45cac795fddd8b95ddb71503L332

Anyway, they are great changes! :)

lehni:
I would be surprised if the math in the different versions of v8 would behave differently, but you never know since we're in edge-case land. The two changes you mention are the only things that actually changed, yes. The `+ 0.5` turns the floor into a round, so I'm wondering if I shouldn't just ouse that instead. Since these are scaling / normalization factors for increased precision, both variants should do the trick and I'm not sure what the reasoning between each is, I just realized I can use the same and it's fine.

lehni:
I was also wondering about `abs(a) + abs(B) + abs(c)` VS `Math.max(abs(a), abs(b), abs(c), abs(d))` in the calculation of the values that get passed to `getNormalizationFactor()`. Shouldn't both use the same approach? Adding them up might be better, as it means we consider the small and large values in our scaling... @hkrish what do you think?

hkrish:
@lehni Both are different ways to normalise and both are valid. 
It really depends on the “condition” of the polynomial representation.
We do consider small and large values while normalisation. If you look at the way it works,

Our norm, (infinity norm — Math.max(sum(abs(a)…)), or L1 norm — sum(abs(a)…)) 
is >1, say 1000 for a set of coefficients
factor ← floor( -log₂(norm) …) = -9.9
this means that, we would scale all coeff. by (2^factor) => 0.0009765625

If norm is less than 1. say 1e-3
factor ← floor( -log₂(norm) …) = 9.9
So in this case we will scale everything up by 1024

It’s really a binary relation; norm <1 or norm >1 are the only cases that we will do something. 
And both norms will arrive at similar values, except for polynomials with coefficients closer to 1. And in those cases, it doesn’t really matter since the normalisation will be vary small, if any at all. (log₂(1.0) = 0).

So both norms shouldn’t make much of a difference as long as we don’t have an order of magnitude of difference between the coefficients (say on the order of 10^7 etc. for doubles). In these cases normalising may not give us any advantage.
And none of these will immediately contribute to any roundoff error since we will not touch a single bit in the mantissa of the coefficients.

I was wondering, are these (root finding code) the main cause of these errors?

lehni:
@hkrish thanks for the explanation! I don't think they're the cause of any errors. I was just wondering if more code could be reused. If you had to choose one approach or the other, which one would you favor?

lehni:
@hkrish another question:

Instead of calculating the discriminant in `solveQuadratic()` and using it to determine if the coefficients should be normalized, could we also do this there, just as in `solveCubic()`?

```js
solveQuadratic: function(a, b, c, roots, min, max) {
    var s = Math.max(abs(a), abs(b), abs(c));
    // Normalize coefficients à la Jenkins & Traub's RPOLY
    if (s < 1e-8 || s > 1e8) {
        // Scale the coefficients by a multiple of the exponent of
        // coefficients so that all the bits in the mantissa are
        // preserved.
        var f = getNormalizationFactor(s);
        a *= f;
        b *= f;
        c *= f;
        B *= f;
    }
   ...
}
```

hkrish:
If I had to choose a norm, I will choose infinity norm (Max). (since it will be smaller than L1norm)
Like I said, it might not matter in most cases. So if you have something in mind feel free to choose.

Actually checking if determinant is very small is a better test, since we are making decisions based on discriminant in the quad case and might miss roots.
For cubic that is not possible/necessary, since we are not calculating (and making decisions) based on discriminants anyway.

p.s. the test (s < 1e-8 || s > 1e8) is just a performance optimisation. Since we usually don’t have to bother normalising within some range closer to 1 (1e8 is an arbitrary choice from my side, if we have some data, we can choose differently or omit it all together).
Similarly if discriminant is well above or below 0, we don’t have to really normalise. In most cases for quadratic, the normalisation may not take place. For extreme situations, dealing with very small curves etc., the main issues might be somewhere else.

lehni:
Thanks @hkrish, I've incorporated this in 9d6aab3802082439f0fe933a22fddf14dfe34173 now. Works like a charm.

PR #1092, opened 2016-06-22
Accidental semicolon in var declaration

iconexperience:
I think the semicolon should be a comma, because `modifiers` is not declared anywhere else.


lehni:
You're right, thanks!

Issue #1109, opened 2016-07-17
Unite operation wrongly fills inner path 2


iconexperience:
Here is another example that still fails:

![image](https://cloud.githubusercontent.com/assets/2122773/16902817/aea771bc-4c6a-11e6-8d97-0d613f7fbfce.png)

The shape in the middle should not be filled.

Here is the code:

```
var p1 = new paper.CompoundPath();
p1.addChild(new paper.Path.Rectangle(200, 100, 200, 200));
p1.addChild(new paper.Path({segments:[[274.616046379547, 215.30654471841746, 0, 0, 0, 0], [296.47810618021146, 202.68940186127458, 0, 0, 0, 0], [297.0266776087829, 160.32607960213838, 0, 0, 0, 0], [260.5870762798793, 181.99677727655697, 0, 0, 0, 0], [260.5870762798793, 207.20554804067655, 0, 0, 0, 0]], closed:true}));
var p2 = new paper.CompoundPath();
p2.addChild(new paper.Path.Rectangle(210, 110, 180, 180));
p2.addChild(new paper.Path({segments:[[260.5870762798793, 207.2055480406766, 0, 0, 0, 0], [274.616046379547, 215.30654471841748, 0, 0, 0, 0], [296.47810618021146, 202.6894018612746, 0, 0, 0, 0], [297.0266776087829, 160.3260796021384, 0, 0, 0, 0], [260.5870762798793, 181.996777276557, 0, 0, 0, 0]], closed:true}));

p1.strokeColor = "green";
p2.strokeColor = "red";
var res = p1.unite(p2);
res.fillColor = "rgba(0, 0, 255, 0.2)";
```

[...]

PR #1112, opened 2016-07-20
Change winding calculation if point on path

iconexperience:
Add a winding if the point is on the path and windings canceled each other. See discussion at #1109


lehni:
Thanks, @iconexperience!

PR #1132, opened 2016-08-06
Set layer._parent in insertLayer

sapics:
It would be necessary to set layer._parent for layer.owner is still set, if layer._index set to undefined.
https://github.com/paperjs/paper.js/blob/develop/src/item/Item.js#L2552

I guess that it would fix problem found in https://groups.google.com/forum/#!topic/paperjs/rAp0uyj5BcU


sapics:
It would be error in layer.isInserted https://github.com/paperjs/paper.js/blob/develop/src/item/Layer.js#L78.
I will reopen after fix it.

PR #1155, opened 2016-08-31
Use correct value for y 45 degree tangent

iconexperience:
I got the pythagoras wrong when I wrote this code. The length of the tangent vector is 1, therefore the y component is sqrt(0.5) for 45°. This will not make any difference, but still we should use the correct value.

image

### DIFF
--- a/src/path/PathItem.Boolean.js
+++ b/src/path/PathItem.Boolean.js
@@ -611,10 +611,9 @@ PathItem.inject(new function() {
                     // Determine the direction in which to check the winding
                     // from the point (horizontal or vertical), based on the
                     // curve's direction at that point. If the tangent is less
-                    // than 45° (y of tangent < sqrt(0.5)), cast the ray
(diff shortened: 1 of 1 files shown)


lehni:
Yes that make sense. We could use this instead: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/SQRT1_2

iconexperience:
I did not even know that existed. It's definitely better, even though precision is not important here.

lehni:
Precision isn't, but understanding is better : )
We could also do `new Point({ angle: 45, length: 1 }).y` : )

iconexperience:
Uhm, yes, we could do that...

iconexperience:
Closed and reopened this pull request to trigger a Travis-CI rebuild. This time all checks have passed.

lehni:
Thanks! : )

PR #1156, opened 2016-09-06
Add try-catch block to handle error on _contains

kylemeenehan:
Added a try-catch block to return false by default on a containment check. This is to fix an error thrown when checking mouse over or click on an irregular shaped group.

I encountered an error when adding hover and click events to groups. The error thrown was due to `point` being null.

I saw this comment in the source though:

`// We only implement it here for items with rectangular content,
        // for anything else we need to override #contains()`

So I'm not sure whether there is an existing way to handle this error?

Loving paper! Just finishing up a big project built largely using paper.

### DIFF
--- a/src/item/Item.js
+++ b/src/item/Item.js
@@ -1688,7 +1688,11 @@ new function() { // Injection scope for various item event handlers
         }
         // We only implement it here for items with rectangular content,
         // for anything else we need to override #contains()
-        return point.isInside(this.getInternalBounds());
+        try{
+            return point.isInside(this.getInternalBounds());
+        } catch (e){
+            return false;
+        }
     },
 
     // DOCS:


lehni:
Instead of adding exception handling, I think we should try and solve the origin of the problem. Could you provide a test-case where the error occurs? Thanks!

kylemeenehan:
Good plan! I'll try build a pen that demonstrates the error.

The project will be publicly viewable from the 19th of September, I'll definitely send a link your way!

lehni:
Great, thanks!

lehni:
I hope the launch went well. Please share the pen when you get to it! thanks!

kylemeenehan:
Sorry about the delay, still in chaos with this project, haha! I've tried to isolate the error unsuccessfully. We've got a new launch date which is the 3rd of October and after that I'm going to try isolate this error.

We've got a staging link at the moment for desktop (chrome). If you want to see the paper.js part, click on the menu button that says "change the 24 hour experience" after the intro video:

http://24hoursstage.azurewebsites.net/

I'm using my version with the error caught, I'd like to work on isolating the error, but if I can't then I'll send you a link where the current version of paper.js is used.

kylemeenehan:
I haven't had time to work on this unfortunately. I can close this PR to clean up your open PR's if you're happy that PR #1651 addresses this?

PR #1175, opened 2016-09-25
Optional clearing of canvas before redrawing

Nuuf:
Adds an option for disabling clearing of canvas before redrawing


lehni:
Could you describe a context in which this makes sense?

Nuuf:
I dont believe it has a massive range of practical uses and you could easily just implement it yourself by overriding the method but I thought it would be nice to have it integrated.

The reason I use it is for creating a few shapes and then draw with them. This has however little to no uses for SVGs or such. It depends on how Paper.js is used.

https://gyazo.com/2191a237c0e8d9027dbfb24e88c67621 Sample image

I just started using user Paper.js and I'm not aware of its full capacity yet but I really enjoy it :)

PR #1184, opened 2016-10-21
Remove extraneous check in importJSON

noahbass:
This check in importJSON seems unnecessary... `layer && layer.isEmpty() && layer`.


sapics:
The second argument is `target`.
https://github.com/paperjs/paper.js/blob/develop/src/core/Base.js#L536

I guess that it looks necessary.

lehni:
@noahbass it is necessary. It means that the `layer` object is returns if it is defined and empty, otherwise `undefined`. Similar to this, but shorter:

```js
return layer && layer.isEmpty() ? layer : null
```

PR #1193, opened 2016-11-04
Fix two issues in PathItem.compare()

iconexperience:
Variable `count` was never initialized, so incrementing resulted in a NaN.
Creating `paths2` failed if `path._children` was `undefined`.


lehni:
Seems legit! Will merge right away.

PR #1217, opened 2016-12-10
Fix example in jsdoc

### LINKED ISSUE #1211: Documentation - http://paperjs.org/reference/item/#parent second example wrong
URL: http://paperjs.org/reference/item/#parent

The second example is wrong as when run in sketch the second two logs output false.  I think it was meant to be...

```
var path = new Path();

// New items are placed in the active layer:
console.log(path.parent == project.activeLayer); // true

var group = new Group();
**path.parent = group;**

// Now the parent of the path has become the group:
console.log(path.parent == group); // true

// The path is now contained in the children list of group:
console.log(group.children[0] == path); // true
```

sapics:
Close #1211

### DIFF
--- a/src/item/Item.js
+++ b/src/item/Item.js
@@ -1276,7 +1276,7 @@ new function() { // Injection scope for various item event handlers
      * console.log(path.parent == project.activeLayer); // true
      *
      * var group = new Group();
-     * group.parent = path;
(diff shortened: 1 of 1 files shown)


lehni:
Thanks @sapics!

PR #1246, opened 2017-01-26
Bug fix SvgExport.js

jmostaer:
Missing arguments in function call to getDefinitions

### DIFF
--- a/src/svg/SvgExport.js
+++ b/src/svg/SvgExport.js
@@ -334,7 +334,7 @@ new function() {
         // Make sure the definitions lookup is created before we use it.
         // This is required by 'clip', where getDefinition() is not called.
         if (!definitions)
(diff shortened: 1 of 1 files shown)


lehni:
What bug is this fixing? Could you tell me a scenario / provide a test-case where this is causing issues?

jmostaer:
If getDefinition is called without arguments (which sometimes happens when calling paper.project.importSVG) this line inside the getDefinition function will throw an exception since item is undefined:

var id = item._id || item.__id || (item.__id = UID.get('svg'));

lehni:
Which version are you using? I believe we already fixed that bug in the `develop` branch.

jmostaer:
I'm using the version on the master branch. Develop seems to be fixed indeed when looking at the code differences.

PR #1253, opened 2017-02-07
Add Line.getDistance() and use it in Curve.getOverlaps()

iconexperience:
Added a static function `getDistance()` to line, that simply calls `Line.getSignedDistance()` and returns the absolute value. This makes the `Line` API a bit more symmetrical, which already has the mathod `getDistance` for the `Line` prototype.

In Curve.getOverlaps() we create a new instance of `Line` and at least one instance of `Point` to check if the two curves are overlapping lines. This is done for each pair of curves and the result is false in the vast majority of cases. Since this generates quite a lot of garbage for little effect, so I though it would be nice to implement a way that does the same without producing garbage. Also, the new code is smaller when minified, so this compensates a little bit for the extra code in `Line`.

### DIFF
--- a/src/basic/Line.js
+++ b/src/basic/Line.js
@@ -182,6 +182,10 @@ var Line = Base.extend(/** @lends Line# */{
             return ccw < 0 ? -1 : ccw > 0 ? 1 : 0;
         },
 
+        getDistance: function(px, py, vx, vy, x, y, asVector) {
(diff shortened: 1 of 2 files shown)


lehni:
Yes this makes a lot of sense. Thanks for spotting! We may want to use this approach in other parts too, e.g. in the code that calculates `strokeBounds`...

PR #1258, opened 2017-02-21
Fix svg export

mazhekin:
Hi Lehni, I have downloaded your new version 0.10.2. But all of the sudden I stuck with an error in svg export function. I fixed it locally in my repository and the problem disappeared. So fix it please, if my proposition is the correct


lehni:
This is already fixed in the `develop` branch.

mazhekin:
Ok, thanks, when it will be in npm version? becouse as I install via npm inspall papep --save the bug still reproduced.

Issue #1261, opened 2017-02-22
Unite operation results in empty path


iconexperience:
Here is an example of a unite operation that results in an empty path:

```
var p1 = paper.PathItem.create("M933.13,1023.97l-516.19,-171.71l607.1,-101.57z");
var p2 = paper.PathItem.create("M933.13,1023.97l-516.19,-171.71l67.33,-11.27l337.49,112.27c16.7,-50.22 71.03,-77.36 121.34,-60.63l45.23,-135.96l35.71,-5.98z");
var res = p1.unite(p3);
```

<a href="http://sketch.paperjs.org/#S/pVHBSsQwFPyVktMupI+8lybZrnjy5EEQPBoP2W7cLc22Ja0eFP/dl8WD4FESHpnMMMwkn2IMlyj24mmIa3cWUnTTseD3kKsZq9tqDnPM8BjW8/0aL9DlGNa48eKh1RpQS1SkoXWpNmgBW1mjQ3CYrHKAjBSCcR9ebG/8eDWl/5g60JoRArmktYOmlYjEqGOhk7VRQFSxWLHMsdpWSAi6kbVVYHVqDFBx0AZam3g6Dsnn3U/EGWFZ8zTEuylNmaN6kePRi0LRH+qUYxyvZKmW41K6IbyNPdeZqRjyJbz2Kf3yOx3ChoyR1XUo3kBbL/xYFn/BgZ9jmKd+XBexf375+gY=">Here is the sketch</a>, but note that in the sketch everything works as expected.

![image](https://cloud.githubusercontent.com/assets/2122773/23228978/e46c92f6-f93e-11e6-82e3-f40d43944b91.png)

lehni:
Ok, thanks. 

[...]

PR #1264, opened 2017-03-01
Fix addWinding and improve performance slightly

iconexperience:
Fixes the handling of special cases in addWinding(). Cases where the curve is crossed by the horizontal/vertical ray at the start point, have to be treated in a special way in order to prevent double counting of crossings (at end point of previous curve and at start point of current curve). This was not handled correctly before.

Also, if a curve is completely right/left of the specified abscissa, we now select 1 as the curve time. Any value (except 0, which requires special handling) can be used for the curve time, and 1 does not require any calculations for the point (it is simply the end point of the curve), therefore this could improve performance slightly.


iconexperience:
Details on the addWinding() fix can fe found <a href="https://github.com/paperjs/paper.js/issues/1261#issuecomment-282726147">in this comment to #1261</a>

PR #1298, opened 2017-04-03
Fix item.transform

sapics:
By the commit https://github.com/paperjs/paper.js/commit/3ee46ffc5c85784a34f99c9d56e8b1e04c038189, decompose.shearing is renamed to decompose.skewing.

### DIFF
--- a/src/basic/Matrix.js
+++ b/src/basic/Matrix.js
@@ -671,7 +671,7 @@ var Matrix = Base.extend(/** @lends Matrix# */{
 
     /**
      * Attempts to decompose the affine transformation described by this matrix
-     * into `scaling`, `rotation` and `shearing`, and returns an object with
+     * into `scaling`, `rotation` and `skewing`, and returns an object with
      * these properties if it succeeded, `null` otherwise.
      *
      * @return {Object} the decomposed matrix, or `null` if decomposition is not
--- a/src/item/Item.js
+++ b/src/item/Item.js
@@ -473,7 +473,7 @@ new function() { // Injection scope for various item event handlers
      *     light', 'color-dodge', 'color-burn', 'darken', 'lighten',
      *     'difference', 'exclusion', 'hue', 'saturation', 'luminosity',
      *     'color', 'add', 'subtract', 'average', 'pin-light', 'negation',
-     *     'source- over', 'source-in', 'source-out', 'source-atop',
+     *     'source-over', 'source-in', 'source-out', 'source-atop',
      *     'destination-over', 'destination-in', 'destination-out',
      *     'destination-atop', 'lighter', 'darker', 'copy', 'xor'
      * @default 'normal'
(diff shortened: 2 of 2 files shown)


lehni:
@sapics looks interesting, but could you explain this a little? Also, it doesn't look like PR includes 3ee46ffc5c85784a34f99c9d56e8b1e04c038189 ?

sapics:
Sorry for less explanation and bad English.

By the commit https://github.com/paperjs/paper.js/commit/3ee46ffc5c85784a34f99c9d56e8b1e04c038189, `var decomp = bounds && matrix && matrix.decompose();` return
```js
{ translation: this.getTranslation(),
  rotation: rotate * degrees,
  scaling: new Point(scale),
  skewing: new Point(skew[0] * degrees, skew[1] * degrees)};
```
from https://github.com/paperjs/paper.js/commit/3ee46ffc5c85784a34f99c9d56e8b1e04c038189#diff-98ece3b8f6303f8a0f4da3c45017b814R683.
Therefore, `decomp.shearing` is always `undefined`.
To check the matrix has shearing or not, we need to check `decomp.skewing.isZero()`.

lehni:
Haha, now I get it! I thought 3ee46ffc5c85784a34f99c9d56e8b1e04c038189 was yours too, forgot that I wrote it : )

lehni:
We should create a test-case for the SVG export scenario also.

lehni:
Could you describe a scenario where the old behavior is causing issues? I would like to create a test-case for it to prevent regressions.

sapics:
@lehni I have tried to create a test-case, but I could not.
I guess that `var decomp = bounds && matrix` is always `undefined` or `null`, from below patterns.

When `matrix` is `undefined` or Identity Matrix, `decomp` is `undefined` or `null` from https://github.com/paperjs/paper.js/blob/develop/src/item/Item.js#L3381

When `matrix` is Non-Identity Matrix, then `transform` is `true` in https://github.com/paperjs/paper.js/blob/develop/src/item/Item.js#L3384.
Thus, `_matrix.prepend(matrix)` run in https://github.com/paperjs/paper.js/blob/develop/src/item/Item.js#L3401
and `_matrix._changed()` in https://github.com/paperjs/paper.js/blob/develop/src/basic/Matrix.js#L430
And `_matrix` has owner, then, `owner.transform(null, true);` or `owner._changed(/*#=*/Change.GEOMETRY);` run in https://github.com/paperjs/paper.js/blob/develop/src/basic/Matrix.js#L129.
If `owner._changed(/*#=*/Change.GEOMETRY);` run, `owner._bounds` is undefined, thus, `var decomp = bounds && matrix` is `undefined`.
If `owner.transform(null, true);` run, then `owner._changed(/*#=*/Change.GEOMETRY);` run in https://github.com/paperjs/paper.js/blob/develop/src/item/Item.js#L3447, thus `var decomp = bounds && matrix` is `undefined`.

If there are cases where `item._matrix._owner === item` is false, there are some possibilities that `var decomp = bounds && matrix` is not `undefined`, `null`.
But, I guess that `item._matrix._owner === item` should be true.

sapics:
Thanks @lehni! This was fixed in v0.11.3!

lehni:
@sapics I was thinking if your detailed analysis when I worked on this, but couldn't find it in the issues. It all worked out in the end :)

PR #1339, opened 2017-06-13
start index can be negative in Path._add() if only one segment has been added to the path

stity:
The following (simplified) code was producing the error : "cannot set property _path of undefined" :
```javascript
path.removeSegments();
path.add(anotherPath.segments[0].point);
```

After a bit of search, it appeared that the `start` index in `_add` was negative, this lead `curves[i]` to be `undefined`

The proposed change should fix this issue.


stity:
[Fixed with the PR #816](https://github.com/paperjs/paper.js/pull/816)

PR #1358, opened 2017-07-21
Updates key.js deal with auto-filling from Chrome

cjds:
This deals with the auto-filling from chrome which registers as an undefined key because of the amount of characters that this enters. 

If you don't have this in this error occurs on text-boxes that are auto-filled

https://www.screencast.com/t/8BdlCBMgY


lehni:
@cjds thanks for the fix, and apologies for taking so long to merge it!

PR #1372, opened 2017-09-02
Fix broken markdown links

alexjlockwood:


Jolg42:
Duplicate of #1417

lehni:
Thanks @alexjlockwood  and @Jolg42 for looking into this. I actually prefer this version because of the way the lines break after the links.

PR #1375, opened 2017-09-05
Fix #1302. By using an empty string as fallback.

### LINKED ISSUE #1302: Errors with browser autofill
When I open a website using paperjs using a browser with autofill enabled, paperjs-full keeps throwing error that `key.length is undefined`. [This line](https://github.com/paperjs/paper.js/blob/6e7fe67a8ddc93d41b620ef8f02a503d2920769b/src/event/Key.js#L88) causes the problem - when the browser autofills, the event is a `keyDown` event, but there is no `event.key` causing this to fail.

jo-soft:


jo-soft:
unfortunately I was unable to provide some unit tests for this fix.

jakubzitny:
Are you planning on fixing the tests here, please, @jo-soft?

jo-soft:
the failing tests are unrelated to my fix and also fail if i set my head back to c403c86a23f54f07bb7e9448c7f05df5cebb464c. (two commits before my fix on my branch)
I had a look at the failing tests but I do not have enough knowledge  about the system to fix then. I could do the simple fix, just adjusting numbers and images to make it work again but this would mean i stil dont know anything why they are failing.

If somebody gives me a hint i'll try to have a look at them again.

jo-soft:
closed because related issue #1302 is included

Issue #1186, opened 2016-10-24
Configurable selection stroke width


mladen107:
Hello
Could we make this not hard coded and put it in settings?

https://github.com/paperjs/paper.js/blob/develop/src/item/Project.js#L873

lehni:
This is the same as #433, which has been there for a long time, but is not high priority for now, as there are many other, more pressing issues for `v1.0.0`.

Issue #433, opened 2014-03-25
Feature Request: Provide Item.selectedStyle to customize selected items


Duffycola:
Currently, there is a color property (Style.selectedColor) that can be set to customize how a selected item is rendered. However, it would be great to have more options. In particular I require to set
- strokeColor, fillColor (or more general the style)
- opacity

So in my view selectedColor should be replaced with a property selectedStyle of type Style, which would cover up strokeColor and fillColor.

I would also request opacity to move into the Style class.

On a side note, I would also like to customize opacity for fillColor and strokeColor individually, because I would like to have a transparent fill, but a non-transparent stroke.

Attached you can find three examples of different item states where I want items to be rendered differently (default, selected, editing).

![bildschirmfoto 2014-04-06 um 13 25 59](https://cloud.githubusercontent.com/assets/1650328/2625119/6f158d72-bd7e-11e3-8654-49fd1e2f2269.png)

nomfecris:
I have been looking at how to do this, as I am wanting to use it on another project, and I wouldn't mind starting a fork and contribute back. But just checking if this is maybe already being looked into further (as last communication is over a year back).

Let me know your thoughts.

lehni:
This isn't being worked on currently, but I agree that we need this, and other people think so too, see #1186. If anybody wants to start working on it, let's use this issue to discuss a viable approach first.

rudrakpatra:
I want to select some of the segments of a path.
segments.selected=true
is very nice 
but
I think something like
segment.selectedColor="orange"
segment.anchorSize=2// visual size of the knob

PR #1380, opened 2017-09-15
Add drawSelectionHandles and selectionLineWidth options to paper.settings

fsih:
I'm writing a paint editor for a kids' site. For my project, I need to change the look of selected items. Are you guys open to adding more options to paper.settings to customize the look of the select state? I'm happy to discuss other options.

drawSelectionHandles: Allow this to be set to false to not draw the control points and handles on selected items. They show the blue outline only.

selectionLineWidth: Allows this to be set to > 1 so that the blue outline is more defined.

### DIFF
--- a/src/core/PaperScope.js
+++ b/src/core/PaperScope.js
@@ -52,7 +52,10 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
             applyMatrix: true,
             insertItems: true,
             handleSize: 4,
-            hitTolerance: 0
+            hitTolerance: 0,
+            drawSelectionHandles: true,
+            selectionLineWidth: 1
+
         });
         this.project = null;
         this.projects = [];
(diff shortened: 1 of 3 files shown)


lehni:
I think something like this makes sense indeed, and there have been related PRs and discussions in the past:

- #433
- #1186

But before we settle on a solution, there are some things to consider:

- Shouldn't this be part of the `Style` class really? Unfortunately, currently styles don't cascade in Paper.js (see https://groups.google.com/forum/#!msg/paperjs/pJzsEYv2W7M/h-jSWu3j-tQJ for a discussion of a planned change). Once this is in place, I think we should really have it there, with the values in `settings` perhaps still working as a global fallback (or they are fully removed)

- Once there are multiple options, I think it makes sense to group them in a property, e.g. `settings. selectionStyle`. This could even be a subclass of `Style` called `SelectionStyle` with all the properties from there and some additional ones. These already existing properties would make sense in this context: `SelectionStyle#strokeColor`,  `SelectionStyle#fillColor`,  `selectionStyle.strokeWidth`, `SelectionStyle#opacity`. New ones that would be added:  `SelectionStyle#handleSize`,  `selectionStyle.handlesVisible` (to reflect the naming of the `#visible` property)

fsih:
For 1, do you mean you want to wait on submitting this PR until the styles cascading work is done, or that  it's OK to add to settings for now and move to styles later?

For 2, I'm looking at adding a settings.selectionStyle. Do you want to keep handleSize where it is in settings.handleSize to avoid breaking changes?

PR #1382, opened 2017-09-21
Fix path.isSelected

### LINKED ISSUE #1378: Possible copy/paste bug in Curve#isSelected()?
I was browsing the source code and noticed the following code:

https://github.com/paperjs/paper.js/blob/b6f9d73af0c44d9f75c880730465e05f85fb690e/src/path/Curve.js#L342-L347

The code checks if `handle2` is selected twice. Should it be checking if `handle1` is selected as well?

sapics:
This fixes #1378


lehni:
Thanks @sapics!

Issue #1328, opened 2017-05-17
wrong position with svg imported embedded Raster with transform="matrix()"


s-light:
i found a bug that has to do with if an svg element has an `transform="matrix(0.21 0 0 0.21 43.54 43.531)"` in it.

[example at sketch.paperjs.org](http://sketch.paperjs.org/#S/bVPBjpswEP0Vi1NWSuxNGrEpq/TUai/tKVIvpYqMGcAN2JZnErZa7b/Xhm0XkloIg9+bN8/j8UtiZAdJlhxOQKpJlomyZfwXglGjkYXHSQceldeOcpObgCAQk861v79J8vp5X8kWga1Wn8aYDqQJ7wFjShpWAPMgS1ZIdeKc52aQ5EGGtKmRT7TYng1qjzHVRXqGl/qIjSxtT4AUYAM9e/L27BZ3A2lO4Lpz1tPh+9MiNyyOPGmIHGZCqNJwL/taE1e2E7hqdd2QGM38wmOMR5FWaQqyUMXHNK2gLD4U6+0GdpsdrB82D7syha1a3ysQIbF4T3wkH3ZdWd+NG+cBzpPlaEKIPJnTZ/DLm9Uw4NlJUx6a4AkzRv4My3dQGwRPN8vWfLWyzFh1Noq0NWyhCbq7qexogn22DG0H4ZRMzXpNDYvMOU1Zg7YF3tp6MbhmbVCHMrgd2LHqU35c47GLwtkM/Mlp5Mn/yKhkC4vttdBN4pnQ8qoTptGvs2J88d76aTU6QJQ13BTkb0KIAf9YU93xM0xx9arRnEU96O/HG8IvGnpe2LMpkSswBH7oz16bEMKxic071wh4uHFFuBsnZ7UhTLIfP1//AA==)

if i import this file [shadowtest_transformmatrix.svg](https://github.com/s-light/paper.js_tests/blob/6f66eabcb966fedb3b142e828e17278d6e4c10ce/svg/shadowtest_transformmatrix.svg)
i get 
![image](https://cloud.githubusercontent.com/assets/1340319/26149947/a39cbb8e-3afc-11e7-8b6e-d39133a92331.png)

but i would expect
![image](https://cloud.githubusercontent.com/assets/1340319/26149887/6021ab3a-3afc-11e7-8f71-566c23ef16eb.png)

this transform arrangement can happen with illustrator exports (there i originally found it)
[...]

Issue #1192, opened 2016-11-02
boundary of raster image element inside svg


hishin:
I am loading an svg image onto paper using:

importSVG(mysrc.svg)

The svg includes text as well as raster images within it.
For example, mysvgitem.children[0] is a PointText item, and mysvgitem.children[1] = Raster item.
The svg loads correctly, but when I try to get the boundary information of the raster element inside it for instance

mysvgitem.children[1].boundary

it gives me completely different numbers for width, height and x, y position (than what would show up as the actual boundary, e.g. when I do mysvgitem.children[1].selected = true)

Is this a known bug?

supernlogn:
This happens also when you load an SVG image directly to a Raster Element.
Example:
`raster = new paper.Raster('mySvg.svg')`
Now if i print the rasted dimensions( `raster.width`, `raster.height`),
The printed dimensions are obviously wrong.

But I found a way around this bug:
Lets say we have a Raster: `raster = new Raster('mySvg.svg')`
or we import an svg: `raster.importSVG('mySvg.svg')`

Then `raster.width` holds the value we want this raster element to have
Unfirtunatelly, setting `raster.width = raster.width` or `raster.setWidth(raster.getWidth()`
has no effect.

The way of fixing it is to write:
`raster.setHeight(raster.height+1);
 raster.setHeight(raster/height-1);`

Ofcourse the svg image presented has not the initial quality, But at least it is fixed
That is just for now. 
I hope the developers will solve this bug soon.

lehni:
Could you share the original SVG file that produces this scenario? Hard to look into an issue without seeing the data. Thanks!

lehni:
@supernlogn I am still hoping to get some SVG data to reproduce this. Could you share it?

supernlogn:
Here is the image: https://drive.google.com/file/d/0BxzY0txWsh_MN2x0TTRKc2xzTTg/view?usp=sharing
I am just 4 months late :P
Try this:
`
raster = new paper.Raster("theImageIuploaded.svg"); // create new raster from an SVG
raster.translte( paper.Point(300,300)); //center it
raster.setHeight(raster.height+1); raster.setHeight(raster.height-1); // Bug ahoy!!
`
To see that there is a bug. The image suddently shrinks to 151 pixels.

lilltiger:
I am in badly need for this stuff to work, so I wonder if there is any progress at all on this front.

[...]

Issue #1348, opened 2017-06-20
Access SVG id attributes after import


kontur:
I would like to identify some groups in an imported SVG based on their `id` attributes (generated by Illustrator's export based on layer names). I can't find a way to retrieve those in the imported object - do they get dropped completely, or are they somewhere there under the hood still?

#745 was the same question / request, but closed without resolution.

lehni:
It ends up in `item.name`

nicopowa:
Here's my experiment, SVG file imported with elements IDs.
I'm able to get Paper items based on child SVG ID, pretty useful to change fill or stroke.
https://github.com/nicopowa/PaperSVGLoader

PR #1452, opened 2018-03-01
fix issue with svg import of raster images

park9140:
### Description
When importing rasters after load if the image has a matrix with any translation/skew value
the image will calculate the wrong matrix after loading.

In order to correct this we append the translation rather than prepend so
that it happens in the correct order relative to the transformation.

#### Related issues

- Relates to #1192
- Relates to #1328 

### Checklist

- [ x ] New tests added or existing tests modified to cover all changes

### DIFF
--- a/src/svg/SvgImport.js
+++ b/src/svg/SvgImport.js
@@ -222,9 +222,8 @@ new function() {
                 // half of its size. We also need to take the raster's matrix
                 // into account, which will be defined by the time the load
                 // event is called.
-                var center = this._matrix._transformPoint(
-                        getPoint(node).add(size.divide(2)));
-                this.translate(center);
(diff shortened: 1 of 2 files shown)


park9140:
Tests on this would pass if develop branch were passing.

Would really love to have an npm package update with this fix.

lilltiger:
Awsome! I am egaerly waiting for this patch, right now I am using ugly hacks to circumvent the issue.

But how does #1348 relate to this issue?

park9140:
@lilltiger  Oh good catch, I typed 1348 rather than #1328. Fixed.   I have a copy of this built in a branch on my fork that i'm installing from for now.

lehni:
LGTM, thanks @park9140 !

Issue #1457, opened 2018-03-09
Path#intersects() returns a wrong result under a specific condition


kota65535:
# Description/Steps to reproduce
Here is the reproduction test code, that also explain about the specific condition.
http://sketch.paperjs.org/#S/pVRNT8MwDP0rVi9tpTHailMlThw4ISHgBhyy1EujhbhKXMaH+O8kK+vKGAJEK0Wp/Gy/Zz/1NbHiAZM6uV4hyzaZJZKa+A3hOT6GmxZh0SsgKXvngax5hnWLFkTXmecLwU4/1bAUxuP8zsak4Qyp0qFgBAGd4BbWOhydfiQGj8zaqgFokKEr4RQsruEyIOdXKFlYZTB7HSAw5NUDhLTlrCyKWZHPxjh5zZrsFPIJ4PULDsHrcMuqkB5K7OJLbcwZGXI1pMoh2nQMfRU6RN7yOztq9VHFII6AW8HhQPD9A9Byc5WtNk3qP0DCNqDZo1mC9hCoQpFvJxcHorYDOXfUd7s5bKo4DDpvu/J+9u14jn41H5ggfqVy3KiKtDbSglr0DKEgOh8WF5pMZFQ/ylBTGQdYxk1VxZ+Ixs7uJ0dJjIz3LQWfWu2Zptw3jWdHK9zaxmGT7qiMRKo/Ezn5J5GFEXKVHlgft8FsvqXeNLDA6fAkWU8G54ZU5sr5uE2fqSrPvyvArj+UX33Nj2/4tSyCf1ZdFOmT+vb+7R0=

# Expected result
Path#intersects() works as normally.

# The cause
I managed to find the cause of this bug.
This is because of the wrong use of cache in Item#_getCachedBounds().
- path `p1` has matrix with `tx=-100, ty=0`
- group `g1` has matrix with `tx=100, ty=0`
- To calculate bounds of `g1`, `Item._getBounds()` are called
- It calls `item._getCachedBounds()` with matrix, which is the sum of `g1` and `p1`'s matrix, resulting in **identity matrix**.
- In `item._getCachedBounds()`, if the `matrix` argument is an identity matrix, internally stored matrix (`p1`'s matrix, `tx=-100, ty=0`) is used to calculate bounds instead of the given identity matrix. This causes the bound wrongly placed by -100 in x-axis.

PR #1458, opened 2018-03-09
fix intersection bug

kota65535:
Fix intersection bug under the following condition (#1457)
- multiplying parent and child matrix results in identity matrix
- applyMatrix is false

### DIFF
--- a/src/item/Item.js
+++ b/src/item/Item.js
@@ -905,7 +905,6 @@ new function() { // Injection scope for various item event handlers
         // See if we can cache these bounds. We only cache the bounds
         // transformed with the internally stored _matrix, (the default if no
         // matrix is passed).
-        matrix = matrix && matrix._orNullIfIdentity();
         // Do not transform by the internal matrix for internal, untransformed
         // bounds.
         var internal = options.internal && !noInternal,


kota65535:
There may be better way than simply removing the line. How do you think, @lehni ?

lehni:
@kota65535 just a heads-up that we're still waiting for feedback on the suggestion by @sapics above. Please advise.

lehni:
@kota65535 could you get back to us? Thank you.

kota65535:
I'm sorry I haven't replied for a long time.
I tried @sapics 's fix for my project, but still no luck.
By the way, I thought this PR fixes the position of the bounds, but it is not true. This PR changes the **position of the rectangle itself**.
The below is the part of the snippet from the [reproduction project](http://sketch.paperjs.org/#V/0.12.1/S/pVRNT8MwDP0rVi9tpTHailMlThw4ISHgBhyy1EujhbhKXMaH+O8kK+vKGAJEK0Wp/Gy/Zz/1NbHiAZM6uV4hyzaZJZKa+A3hOT6GmxZh0SsgKXvngax5hnWLFkTXmecLwU4/1bAUxuP8zsak4Qyp0qFgBAGd4BbWOhydfiQGj8zaqgFokKEr4RQsruEyIOdXKFlYZTB7HSAw5NUDhLTlrCyKWZHPxjh5zZrsFPIJ4PULDsHrcMuqkB5K7OJLbcwZGXI1pMoh2nQMfRU6RN7yOztq9VHFII6AW8HhQPD9A9Byc5WtNk3qP0DCNqDZo1mC9hCoQpFvJxcHorYDOXfUd7s5bKo4DDpvu/J+9u14jn41H5ggfqVy3KiKtDbSglr0DKEgOh8WF5pMZFQ/ylBTGQdYxk1VxZ+Ixs7uJ0dJjIz3LQWfWu2Zptw3jWdHK9zaxmGT7qiMRKo/Ezn5J5GFEXKVHlgft8FsvqXeNLDA6fAkWU8G54ZU5sr5uE2fqSrPvyvArj+UX33Nj2/4tSyCf1ZdFOmT+vb+7R0=) of the original issue #1457.

```
    // create a path with pivot setting
    let p1 = new Path.Rectangle({
      pivot: new Point(100,0),
      position: new Point(0,0),
      size: new Size(200,100),
      fillColor: 'green',
      applyMatrix: false
    })

    // set pivot so that the sum of the child's pivot and itself is (0, 0)  
    let g1 = new Group({
      children: [p1],
      pivot: new Point(-100,0),
      position: new Point(0, 0),
      applyMatrix: false
    })

    // create a group of to test intersection
    let g2 = new Group({
      children: [g1],
      position: new Point(200,200),
      applyMatrix: false
    })
```

Before applying this PR, the final position of the green rectangle `p1` is `(300, 200)`. But after applying it, it changes to `(200, 200)`.
Because pivot is set to the center of the children if not specified, `(200, 200)` seems to be the correct position.
Is my understanding correct?

sapics:
@kota65535 Thanks for your response and checking my fix!
I am sorry, it could not help your problem.

@lehni Let me explain my fix.
```js
var cacheKey = cacheItem && (!matrix || matrix.equals(_matrix))
                && this._getBoundsCacheKey(options, internal);
```
I guess that when `matrix` equals to `_matrix`, we can use `this._getBoundsCacheKey(options, internal);`.
When **matrix is null** and **_matrix is not null**, original calculation `(!matrix || matrix.equals(_matrix))` returns true.
In this situation, my calculation `(matrix ? matrix.equals(_matrix) : !_matrix)` returns false which would be expected result, I guess.
The other situation, the result would be same between original calculation and my calculation.

sapics:
Replacing L928
`var res = this._getBounds(matrix || _matrix, options),`
to
`var res = this._getBounds(matrix, options),`
might fix the problem. But I have less confidence.

Identity matrix has similar meaning as null in paperjs, it is the reason why I would not delete L908.

PR #1502, opened 2018-08-29
Fix order of operations in order to properly cache position calculation.

brandondahler:
### Description
In Item#getPosition, we are caching the `this._pivot` value into `this._position` instead of using the result of the ternary operator that uses `this._pivot` as the condition.

### Checklist

- [X] New tests added or existing tests modified to cover all changes
  - Added, but cannot be run on windows, so not actually tested.
- [ ] Code conforms with the ESLint and Prettier rules (`yarn eslint` passes
      and `yarn prettier` has been applied)
  - Unable to run, was not apparent how to do so.


brandondahler:
I was incorrectly attributing an issue to order of precedence -- in this case conditional comes before assignment and therefore parenthesis are not necessary.

PR #1510, opened 2018-09-20
Fix Path.splitAt() documentation example

sasensi:
### Description

Hi, I fixed a little bug in Path.splitAt() second example which breaks the online [documentation](http://paperjs.org/reference/path/#splitat-location).

As this is my first contribution, I would like to tell you how much I love your library that I use a lot.
I've seen that you are looking for help to maintain the project and I will have a lot of free time in future months.
So if I can be of any help, I would be delighted to.


lehni:
Thanks for spotting! And welcome to the project, @sasensi

PR #1519, opened 2018-09-27
Fix:travis remove "Import SVG Image" in test:node

sapics:
### Description

This pr would fix travis error.
I remove "Import SVG Image" test in test:node, because raster onload event looks not fired in test:node.

### DIFF
--- a/gulp/tasks/test.js
+++ b/gulp/tasks/test.js
@@ -21,7 +21,7 @@ gulp.task('test:phantom', ['minify:acorn'], function() {
     return gulp.src('index.html', { cwd: 'test' })
         .pipe(qunits({
             checkGlobals: true,
-            timeout: 20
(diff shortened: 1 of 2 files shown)


lehni:
It used to work... So the proper fix would be to find out why this isn't work in JSDOM, but that's going quite far. We can do this, but we should add a comment about why node is excluded, and that it should be brought back again.

sapics:
Data URL on img.src looks not supported in node-canvas 1.3, it might cause this problem.
Node-canvas says "Support reading data URL on img.src" is added in node-canvas 2.0.0 unreleased version in https://github.com/Automattic/node-canvas/blob/master/CHANGELOG.md#added.

sapics:
I fixed the problem in #1520. Thus, I close this pr.

PR #1522, opened 2018-09-29
Fix travis errors

sapics:
### Description

This pr fixed all travis errors.

I cannot find out why node v6.13 and v6.14 cause error in SymbolDefinition.initialize.
To avoid these version, I specify test version to v6.12 as temporary solution.
I will make an issue about it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,13 @@ language: node_js
 # Follow https://github.com/nodejs/LTS to decide when to remove a version
 node_js:
 - stable
-- '8'
-- '6'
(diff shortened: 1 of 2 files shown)


lehni:
@sapics it's great you fixed these, thank you!

But for the time being, as a kind of warm-up phase, let's proceed like this:

- When you create a PR, please wait for my review before merging it.
- Please don't commit directly to the `develop` branch yet, work through PRs instead.

Thanks!

sapics:
I understand it. I will take care in the future.

PR #1482, opened 2018-06-11
Fix Cannot read property 'parent' of undefined in node

HriBB:
This happens in webpack bundle in node server rendering react application:
```js
TypeError: Cannot read property 'parent' of undefined
    at Object.eval (webpack:///./node_modules/paper/dist/node/self.js?:20:28)
    at eval (webpack:///./node_modules/paper/dist/node/self.js?:62:30)
    at Object../node_modules/paper/dist/node/self.js (/home/bojan/www/climbuddy/client/server/map.bundle.js:34531:1)
    at __webpack_require__ (/home/bojan/www/climbuddy/client/server/map.bundle.js:26:30)
    at Object.eval (webpack:///./node_modules/paper/dist/paper-core.js?:35:16)
    at eval (webpack:///./node_modules/paper/dist/paper-core.js?:14798:3)
    at Object../node_modules/paper/dist/paper-core.js (/home/bojan/www/climbuddy/client/server/map.bundle.js:34553:1)
    at __webpack_require__ (/home/bojan/www/climbuddy/client/server/map.bundle.js:26:30)
    at eval (webpack:///./node_modules/react-paper-bindings/lib/View.js?:15:18)
    at Object../node_modules/react-paper-bindings/lib/View.js (/home/bojan/www/climbuddy/client/server/map.bundle.js:37188:1)
```

#### Related issues

- Relates to <link_to_referenced_issue>

### Checklist

- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the ESLint and Prettier rules (`yarn eslint` passes
      and `yarn prettier` has been applied)

PR #1525, opened 2018-09-29
Revert "Fix Cannot read property 'parent' of undefined in node"

sapics:
Revert paperjs/paper.js#1482


lehni:
@sapics why did you revert this?

And let's test things locally first before merging and undoing things. Let's keep a clean commit history if we can :)

sapics:
Because it was to **master** branch.
I think that **master** branch should change only new version time, isn't it?

Sorry for moving too quickly. I worked here similar to personal repository, it was bad manner.
I will take care in the future.

PR #1505, opened 2018-09-04
Blend modes

SkepticaLee:
### Description
Provides execution for missing blendMode:
pin-light

Adds new blendModes:
minus
divide
divide-source
hard-mix
pegtop-light
linear-light
vivid-light
grain-extract
grain-merge
darken-intensity
lighten-intensity
divide-intensity
multiply-intensity

Provides newer, simpler code for:
color-burn
color-dodge
darken
lighten
difference

#### Related issues

none
### Checklist

- [x ] New tests added or existing tests modified to cover all changes
- [x ] Code conforms with the ESLint and Prettier rules (`yarn eslint` passes
      and `yarn prettier` has been applied)

[paper.zip](https://github.com/paperjs/paper.js/files/2349840/paper.zip)

PR #1509, opened 2018-09-14
Travis: add 20s timeout in test:node task

sapics:
### Description

Now, travis reports error which come from timeout in test:node.
Thus, this pr add 20s to timeout.

PR #1526, opened 2018-09-29
Fix Cannot read property 'parent' of undefined in node

sapics:
This is PR for develop branch version in #1482.

### DIFF
--- a/src/node/self.js
+++ b/src/node/self.js
@@ -17,7 +17,7 @@ var path = require('path');
 // Determine the name by which name the module was required (either 'paper',
 // 'paper-jsdom' or 'paper-jsdom-canvas'), and use this to determine if error
 // exceptions should be thrown or if loading should fail silently.
-var parent = module.parent.parent,
+var parent = module.parent ? module.parent.parent : null,
     requireName = parent && path.basename(path.dirname(parent.filename));
 requireName = /^paper/.test(requireName) ? requireName : 'paper';
 


lehni:
@sapics one thing: Could you name your feature branches a bit more clearly, so in this case:

`sapics:fix-parent-undifined-on-node`

Or something like that?

Thanks!

sapics:
OK, I will take care in the future.

sapics:
If I need create some branch in paperjs, I will confirm to you before.
For fixing travis error, I need to create new branch in paperjs, because my travis and paperjs travis is not working same.
Sorry doing that without confirmation.

lehni:
@sapics that's really super strange about Travis! I guess our tests run on a different machine or something? But this shouldn't make a difference.  Can you describe what the difference is that you observed?

sapics:
Yes, it's really strange for me, too.
When adding just a number of timeout in **my travis**, nodejs **v4** and nodejs v7 was succeed without replacing git to png. https://travis-ci.org/sapics/paper.js/builds/428491541
However, when I create a PR in #1509 with exactly same code, **paperjs travis** was failed in all versions which includes nodejs **v4** as you see.

I cannot find out what makes changes, but I find that #1505 causing some errors in paperjs travis.
It is also strange for me, but you can see errors from a commit log of paperjs:develop branch.
After I close #1505, all commits have green travis.
I guess that #1505 was the PR from **develop** to **master**, it causes red travis.

image

Issue #1513, opened 2018-09-23
boolean subtract fails for semi-circle inscribed in rectangle


renschler:
# Description/Steps to reproduce
I've reproduced the behavior in the sketch linked below.

I have a rectangle, and then I have a semi-circle that is inscribed within the rectangle.

I am trying to boolean subtract the semi-circle from the rectangle. Example 1: When I do so with {trace: true} the operation returns an empty path (this is wrong - ie not the expected behavior). Example 2: When I do so with {trace: false}, the operation returns a single line, the edge of the rectangle that is shared with the bottom of the semi-circle (I'm not sure if this is wrong, the trace setting is confusing to me but I think it's wrong).

Example 1: Subtract with {trace: true} is supposed to leave me with a path containing the green sections, but it's not doing that.
![screen shot 2018-09-23 at 12 24 34 pm](https://user-images.githubusercontent.com/3973468/45930919-b7072d80-bf2b-11e8-9c17-e8811ac511a2.png)

Note: In my sketch example, I take a few steps to get to the Rectangle and Semi-Circle. That is how this issue surfaced in my code.

[...]

Issue #371, opened 2013-12-21
Parametrized stroke outlining / expanding / offsetting 


strandedcity:
I've been following the library for some time, and have started a spare-time project based on it. I've been eagerly interested in the functionality to expand strokes into paths, particularly for bezier curves (where I know the problem is mathematically complex). Has a strategy / timeline for implementation been discussed? I may be able to contribute if new contributors are welcome.

I wasn't sure if the "outline" feature was separate from what I would describe in a CAD world as "offset". Is there some way to get equations for the position of the stroke edge that is not as complex as a full bezier-offset solution? See http://pomax.github.io/bezierinfo/#offsetting for a very detailed explanation of the math involved.

Thanks!

lehni:
@iconexperience I haven't seen your plunker before, it looks really good! If I understand correctly, the main issue that's stopping you is in the boolean operations code, correct?

iconexperience:
@lehni Yep, a fully working boolean code would solve almost all the remaining issues. There are some edge cases that would need a little work elsewhere, but a boolean code that really works would be a dream come true.
I am extremely busy at the moment (a few days before new product release), but if there is something I should test I will try to arrange a little free time. In a few weeks I should be able to contribute again to PaperJS.

lehni:
@iconexperience that's great to hear. There are still real issues remaining in the boolean code unfortunately, mainly #449 and #450. Unfortunately I haven't heard from @hkrish in quite a while, not sure if he's still working on this other approach or not. We could attempt fixing this in the current code-base together? Your code in #648 would probably be of use here?

lehni:
@kuribas I finally wanted to look at your outline of the boolean code posted in https://github.com/paperjs/paper.js/issues/371#issuecomment-59438784, but that page is gone. Has it moved elsewhere? Would you be interested in joining the effort of implementing proper boolean operation code, as well as offsetting in paper.js?

kuribas:
I am currently debugging my sweep line algorithm for set operations on paths.  It's written in literate haskell, and hopefully clear enough to understand the workings.  I'll upload the file when I am finished.  

As for offsetting, I have found an algorithm to quickly find the best parametrization for a least squares fit.  I then use a binary search to find the best nodes to subdivide.  It is still rather slow, so probably not very useful for real-time applications.  In my tests a least squares solution with a simple approximate parametrization and subdividing the curve will give good enough results (which I believe is what you are doing now?)

[...]

PR #1530, opened 2018-10-01
Add Path/Curve#getTimesAtVectorTangent methods

sasensi:
### Description

I implemented this new methods, as discussed [here](https://github.com/paperjs/paper.js/issues/1513#issuecomment-425647595) to make things like extrusion effect easier.
It will also certainly be usefull for path offset implementation.
As commented in the code, I finally used an algorithm described [here](https://stackoverflow.com/a/34837312/7615922) instead of [this one](https://github.com/paperjs/paper.js/issues/371#issuecomment-277541824) because it already handles edge cases like vertical/horizontal vectors.

### Checklist


src/path/Path.js
```diff
@@ -2123,6 +2123,42 @@ var Path = PathItem.extend(/** @lends Path# */{
      * the beginning of the path and {@link Path#length} at the end
      * @return {Number} the normal vector at the given offset
      */
+
+    /**
+     * Calculates path-time parameters where the path is tangent to given vector.
```
lehni: path-time is a misleading concept that doesn't exist elsewhere in the library. Note that curve-time isn't linear, a curve at t = 0.5 is most of the time no the middle of the curve (at half of its length), because there is tension in curves. path-time here is an emulation where length = 1, 0 = 0, and 0.5 = length / 2.  I think it's better to just speak of lengths instead:

`path.getLengthsWithTangent()`

sasensi: Ok, I understand what you say, so I guess it should return a list of lengths instead of a list of "times" between 0 and 1, right ?
And maybe we could speak of offsets instead of lengths as in other path methods ex : [path.getOffsetOf()](http://paperjs.org/reference/path/#getoffsetof-point) ?
`path.getOffsetsWithTangent()`

src/path/Curve.js
```diff
@@ -2289,6 +2247,57 @@ new function() { // Scope for bezier intersection using fat-line clipping
...
+     * is tangential to provided tangent.
+     * Tangents at the start or end are included.
+     *
+     * @param {Number[]} v curve values
+     * @param {Point} point the tangent to which the curve must be tangential
+     * @return {Number[]} at most two curve-time parameters, where the curve is
+     * tangential to the given tangent
+     */
+    function getTimesWithTangent(v, point)
+    {
```
lehni: @sasensi this should be on the same line as `getTimesWithTangent(v, point)` :)

src/path/Path.js
lehni: Yes you're right! That's the right name for it!

src/path/Curve.js
sasensi: Oops, sorry :)

lehni:
@sasensi for future reference: Instead of merging, could you use rebasing so we don't end up with merge commits each time you get the feature branch up to date with the develop branch?

https://www.atlassian.com/git/tutorials/merging-vs-rebasing

sasensi:
@lehni, about rebasing the pull request instead of merging it, Ok, I'll do that next time.
But the reason I did the merge was to trigger travis build to make sure everything is OK before merging the pull request into master. I did it online through github merging interface.
From what I read in this article (https://www.atlassian.com/git/tutorials/merging-vs-rebasing), rebasing once the pull request is done seems to be risky:
> Reviewing a Feature With a Pull Request
If you use pull requests as part of your code review process, you need to avoid using git rebase after creating the pull request. As soon as you make the pull request, other developers will be looking at your commits, which means that it’s a public branch. Re-writing its history will make it impossible for Git and your teammates to track any follow-up commits added to the feature.
Any changes from other developers need to be incorporated with git merge instead of git rebase.
For this reason, it’s usually a good idea to clean up your code with an interactive rebase before submitting your pull request.

How will github pull request interface react if I force push a rebased version of the pull request ?

lehni:
Travis is run every time you push a branch, wether it is rebased or merged doesn't really matter, both result in new commit hashes in the branch.

I changes a few things around after the merge, have a look at f5366fb3cb53bc1ea52e9792e2ec2584c0c4f9c1

Only minor things, mainly to align it with the rest of the code base (and keep lines under 80 chars : )

sasensi:
Ok well I'll try it next time and see by myself.
Thanks for teaching me all those things btw ! 👍

sasensi:
@sapics thanks, I'll consider setting it up on my fork too.
The only drawback I see with that is that you have to push a commit to run tests.
An alternative that allows you to run tests locally on Windows with a closer environment to travis-ci's one is to use `Docker`. I'm actually trying it and I'm pretty close to get everything working.
If you are interested, I can send you, or even as I was saying to @lehni, put somewhere in contributors note, a way to do that.

sapics:
Oh, that's great! Thanks!!
Currently, I have not enough knowledge about Docker,,, so I would like to use it after your `contributors note` :) 

I failed test at very stating poin :_)
`throw new TypeError('"file" argument must be a non-empty string');`
and
`(function (exports, require, module, __filename, __dirname) { ../src/load.js`
`SyntaxError: Unexpected token .`

After I built paper-full.js by gulp, test passes stating point, and I got some errors in TextItem tests which you were reported.

Issue #1416, opened 2017-11-27
importSVG not applying attributes


okitu:
When using importSVG attributes of nodes are not applied.
The basic attributes are applied directly by the importer-functions
but anything beyond that will just be ignored.

Have a look at this fiddle:
https://jsfiddle.net/4awsphbn/

As you can see, the small circle should be inside the pin,m
but because the "transform" attribute is not applied, it stays 
at the original cx/cy position.

This svg is certainly not optimized, but we have to work with imported
icons, and cannot optimize every icon by hand.

https://github.com/paperjs/paper.js/blob/develop/src/svg/SvgImport.js#L523
```javascript
function applyAttributes(item, node, isRoot) {
    if (node.style) {
    ...
```
This conditional never let's anything through.

I could apply the attributes by removing the conditional, and adding
it in the getAttribute function.

https://github.com/paperjs/paper.js/blob/develop/src/svg/SvgImport.js#L496

```javascript
function getAttribute(node, name, styles) {
    // First see if the given attribute is defined.
    var attr = node.attributes[name],
        value = attr && attr.value;
    if (!value &&  node.style) {
```

Now paper imports the svg with all attributes applied.

sapics:
Thanks @okitu for reporting bug and teaching us how to fix!
And sorry for late response.

This looks serious bug in SvgImport.
Thus, I have created PR in #1532.
If you have any further updates on this, please write down here!

PR #1532, opened 2018-10-02
Fix:SvgImport apply attributes to item

### LINKED ISSUE #1242: ImportSVG: Cannot read property 'width' of undefined
Trying to import an SVG passing directly the root node to importSVG will raise an "Uncaught Error: TypeError: Cannot read property 'width' of undefined" error if there's a width specified in %.

Test SVG (Chartist output SVG):

`<svg xmlns:ct="http://gionkunz.github.com/chartist-js/ct" width="100%" height="400px" class="ct-chart-line ct-chart--manydata" style="width: 100%; height: 400px;"> .... </svg>`

When the root node is passed to importNode it try to get the size calling the getSize without the 5th parameter allowPercent:

`rootSize = getSize(node, null, null, true) || paper.getView().getSize();`

But, into getSize, if width is in % and allowPercent false or undefined (like that call) it try to access to rootSize that isn't initialized yet.

`return /%\s*$/.test(value) ? (res / 100) * (allowPercent ? 1 : rootSize[/x|^width/.test(name) ? 'width' : 'height']) : res;`

As soon as possible i'll provide a test case and maybe a fix

sapics:
### Description
This PR is suggested in #1416 by @okitu.
The condition `if (node.style) {` was introduced in [commit](https://github.com/paperjs/paper.js/commit/261fc02c1ddafe294af4849750091b0f055f917f) to fix #1242.
I cannot judge this condition is necessary or not necessary to fix #1242.
As @okitu pointed out, I guess that we need to remove this condition to apply attributes of node which has no style.

### DIFF
--- a/src/svg/SvgImport.js
+++ b/src/svg/SvgImport.js
@@ -496,7 +496,7 @@ new function() {
         // First see if the given attribute is defined.
         var attr = node.attributes[name],
             value = attr && attr.value;
-        if (!value) {
+        if (!value && node.style) {
             // Fallback to using styles. See if there is a style, either set
             // directly on the object or applied to it through CSS rules.
             // We also need to filter out inheritance from their parents.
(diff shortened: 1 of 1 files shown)


lehni:
@sapics thanks for fixing this!

A little tip: If you write the commit messages like this, then the linked issue(s) get closed automatically when merging:

```
Fix bug in how SvgImport applies attributes to items

Closes #1416
```
If there are multiple issues, you do:
```
Closes #1416, closes #XYZ
```

PR #1533, opened 2018-10-02
Fix #1501 errors with event listeners on mobile

### LINKED ISSUE #1501: Some bug in passive event listener
image

# Expected result
passive event listener   Unable to preventDefault. And I think it can remove `event.preventDefault`in touch event in mobile drive。

# Additional information

sasensi:
### Description

This changes fixes #1501 by explicitely declaring document touchstart and touchmove events as not passive and thus allowing them to be prevented later.
Error was linked to a change in chrome behaviour documented [here](https://www.chromestatus.com/features/5093566007214080). More details on the possible solutions are available [here](https://developers.google.com/web/updates/2017/01/scrolling-intervention).

By investigating this issue, I also discovered another linked bug on mobile when you scroll the page and it is still in scroll inertia. By interacting with the canvas, you can be in a situation where the code tries to call `.preventDefault()` on the event while it is not [cancelable](https://developer.mozilla.org/fr/docs/Web/API/Event/cancelable) and this produces an error.
I fixed this bug in the second commit.

#### Related issues

- Fixes #1501

### Checklist

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -23,8 +23,17 @@ var DomEvent = /** @lends DomEvent */{
             for (var type in events) {
                 var func = events[type],
                     parts = type.split(/[\s,]+/g);
-                for (var i = 0, l = parts.length; i < l; i++)
-                    el.addEventListener(parts[i], func, false);
+                for (var i = 0, l = parts.length; i < l; i++) {
+                    var eventName = parts[i];
+                    // For touchstart/touchmove events on document, we need to explicitely
+                    // declare that event is not passive (can be prevented).
+                    // Otherwise chrome browser would ignore event.preventDefault() calls.
+                    // See #1501 and https://www.chromestatus.com/features/5093566007214080
+                    var options = el === document && (eventName === 'touchstart'|| eventName === 'touchmove')
+                                  ? {passive: false}
+                                  : false;
+                    el.addEventListener(eventName, func, options);
+                }
             }
         }
     },
--- a/src/view/View.js
+++ b/src/view/View.js
@@ -1440,7 +1440,11 @@ new function() { // Injection scope for event handling on the browser
             //   which can call `preventDefault()` explicitly or return `false`.
             // - If this is a unhandled mousedown event, but the view or tools
             //   respond to mouseup.
-            if (called && !mouse.move || mouse.down && responds('mouseup'))
+            //
+            // Some events are not cancelable anyway (like during a scroll inertia
+            // on mobile) so trying to prevent default in those case would result
+            // in no effect and an error.
+            if (event.cancelable !== false && (called && !mouse.move || mouse.down && responds('mouseup')))
                 event.preventDefault();
         },
 


src/dom/DomEvent.js
```diff
@@ -23,8 +23,17 @@ var DomEvent = /** @lends DomEvent */{
...
-                    el.addEventListener(parts[i], func, false);
+                for (var i = 0, l = parts.length; i < l; i++) {
+                    var eventName = parts[i];
+                    // For touchstart/touchmove events on document, we need to explicitely
+                    // declare that event is not passive (can be prevented).
+                    // Otherwise chrome browser would ignore event.preventDefault() calls.
+                    // See #1501 and https://www.chromestatus.com/features/5093566007214080
+                    var options = el === document && (eventName === 'touchstart'|| eventName === 'touchmove')
+                                  ? {passive: false}
+                                  : false;
```
lehni: Some code-style remarks, to align with the paper.js code-style:

- One tab-indent before the `?`, not aligned with `=`
- Spaces inside objects: `{ passive: false }`
- The line looks longer than 80 chars, so it should be broken. Maybe `name` instead of `eventName` would help a little?

I should maybe install ESLint and / or Prettier, so we have a more modern linter and less discussions about code-style :)

lehni:
@sasensi this PR broke our unit tests, and I can't figure out why.

Could you try this out on your own machine?

```
npm install
gulp test --branch develop
```

I get a timeout in the browser-based unit tests. It looks like some callbacks aren't firing anymore and it's just eternally waiting now?

sasensi:
@lehni, tests fail on my machine too, sorry I forgot to run node tests, I will remember for the future.
Strange thing is that they passed in `travis`.
I will fix it.

lehni:
Yeah that's super weird. And once I merged it, Travis started failing!

sasensi:
Did you manage making the tests work again on your local machine by reverting changes ?

sasensi:
First failing build happened after #1535 merge https://travis-ci.org/paperjs/paper.js/builds

sasensi:
I'm investigating, for me, it more look like a a dependency problem than with a code change problem.

sasensi:
I don't know if it is relevant but cloning [gulp-qunits](https://github.com/lehni/gulp-qunits) repository and running `npm run test` fail for me too.

lehni:
Let me check. BTW, you can run the test only in the browser by running `gulp test:browser`. That should also shed more light on what's going on?

sasensi:
Yes, for me, browser tests pass without problem.
It seems that the problem is in `gulp-qunits` + `PhantomJS`.

lehni:
What do you mean with this exactly: 

> I don't know if it is relevant but cloning gulp-qunits repository and running npm run test fail for me too.

Where do you run the tests?

sasensi:
I suspect there is a problem with `gulp-qunits` package so I tried to run its unit test to check that it's working.
And it seems that it's not.

lehni:
I started isolating the problematic tests by deactivating all `include()` statements in the following file, and reactivating them one by one:

https://github.com/paperjs/paper.js/blob/develop/test/tests/load.js

The tests pass until I reactivate this line:

```js
/*#*/ include('Path.js');
```

lehni:
I doubt the issue is in `gulp-qunits`, because nothing changed there, and yesterday it was working :)

sasensi:
I was not thinking about a change in `gulp-qunits` code but maybe a change in its dependency ?

lehni:
It's this test here:
```js
test('#1493 Path#add with 1000000 segments', function() {
    var path = new Path();
    for (var i = 0; i < 1000000; i++) {
        path.add(new Point(0, 0));
    }
    path.clone();
    expect(0);
});
```
When I deactivate it, the tests run for me! This is timing out on PhantomJS (which is a discontinued piece of cr*p, to be fair...)

sasensi:
Oh ok, on my machine, even when commenting all the tests, it still fails. But it must be a problem with my environment.

sasensi:
Oh yes indeed, I guess the 1000000 loop added some time processing... so should we remove this test or increase the timeout ?

I'll continue investigating why the tests are failing on my machine.
What's your local environment by the way ?

lehni:
I'm fixing this right now. What error are you seeing when running the tests?

lehni:
I have some problems too, and they seem to be related to using `nvm` and having an old version of the globally installed `gulp` lying around, so my tests end up running with the wrong version of `node`, which doesn't play well with the compiled native code of `node-canvas` in `node_modules`

sasensi:
For me, it seems that PhantomJS is not working.
```
[16:41:34] 'test:phantom' errored after 20 s
[16:41:34] Error in plugin 'gulp-qunits'
Message:
    The specified timeout of 20 seconds has expired. Aborting...
```

sasensi:
I'm trying to test in a docker environment to see if it's different, but I've got problem with native dependencies too.

lehni:
I'm working on macOS 10.13.6, with homebrew up to date. Node 9.11.2 works well for me.

lehni:
I've just pushed an update. With this, the tests pass for me again.

sasensi:
Great work !
For my part, I'll make up a more robust dev environment with tests passing before making other commits.

lehni:
@sasensi I'm happy to help you getting the setup working, if you're encountering problems. What system are you on?

sasensi:
I'm on Windows 10 Pro x64.
I'm actually trying to setup local travis ci build through `Docker` which should be the closest to the online ci environment.

sasensi:
Sure, and apparently solutions are also proposed with the issues so I'll take care of that.

sasensi:
@lehni, I finally got everything working on Windows through Docker, except 2 unit tests concerning text bounds.
https://github.com/paperjs/paper.js/blob/28dec90a880e050c0efdb222ed1f5852c68a09b4/test/tests/TextItem.js#L24
https://github.com/paperjs/paper.js/blob/28dec90a880e050c0efdb222ed1f5852c68a09b4/test/tests/Item_Bounds.js#L103
In both cases, width is `89` instead of `76.25` as expected.
It seems that it is related to font metrics.
I installed Arial in my container, as it is done in travis.
https://github.com/paperjs/paper.js/blob/28dec90a880e050c0efdb222ed1f5852c68a09b4/travis/install-assets.sh#L13-L23
Does this kind of issues already happened when setting up build process or have you got an idea of what this could be related to ?

lehni:
@sasensi it sounds like your docker isn't using the font somehow, as otherwise the metrics should match. What linux are you using for the container? and are you sure you're installing the font in the right way there?

Also, did you try more on Windows?  Any luck with getting `gulp-qunits` to work there?

sasensi:
About `Docker`, I'm using travis provided nodejs image `travisci/ci-garnet:packer-1512502276-986baf0` which is built over ubuntu.
Here is my Dockerfile:
```
FROM travisci/ci-garnet:packer-1512502276-986baf0

#
# INSTALL NATIVE DEPENDENCIES
#
RUN apt-add-repository -y "ppa:ubuntu-toolchain-r/test"
RUN apt-add-repository -y "ppa:git-core/ppa"
RUN apt-get update
RUN apt-get install -y g++-4.9 libgif-dev

#
# INSTALL NODE
#
# Replace shell with bash so we can source files
RUN rm /bin/sh \
    && ln -s /bin/bash /bin/sh

# Install nvm with node and npm
ENV NVM_DIR /usr/local/nvm
ENV NODE_VERSION 8.12.0
RUN curl https://raw.githubusercontent.com/creationix/nvm/v0.30.1/install.sh | bash \
    && source $NVM_DIR/nvm.sh \
    && nvm install $NODE_VERSION

# add node to path
ENV NODE_PATH $NVM_DIR/v$NODE_VERSION/lib/node_modules
ENV PATH      $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH

# install gulp globally
RUN npm install -g gulp-cli

# install fonts
RUN mkdir -p ~/.assets
RUN mkdir -p ~/.fonts
RUN wget http://sourceforge.net/p/tellmatic/git/ci/master/tree/img/arial.ttf?format=raw -O ~/.assets/arial.ttf
RUN cp -p ~/.assets/arial.ttf ~/.fonts
RUN fc-cache -f -v
```
By the way, maybe once I find a definitive way to make this work on windows I should add a note on that in the contributors part ?

About making qunits work on windows, I haven't looked at it yet, but I will soon.

sasensi:
About Docker integration, I finally resolved it by installing additional native dependencies.
I will work with it for a while to see if it is a shareable / descent workflow for Windows developers.

PR #1534, opened 2018-10-02
Fix #1493 Path#add crashes whith 1000000 segments

### LINKED ISSUE #1493: Maximum call stack size exceeded when cloning a path with many segments

 1. Create a path with more than 125122 segments
 2. Clone it

http://sketch.paperjs.org/#S/VY7BDoIwEER/ZdPEAIEQwXgBOfkDJh48iIdCV9sUuwSqHAj/bku8sKfZmczbnZnhb2QFu2q0rWQJa0n4vUMLPbcSKjA4wcXJMKqNt9LRDqTxTB0NLg6ajrc6KDfhTYm1mzn7SUPocaraQwnqlOXHLD94GccRzLUBN2uXCxGq3VfhlDb0MWJMJ89JYGtKVC9pI4de/kfbjgyGUen+bwbkuidl7MiK+2P5AQ==

# Note

It is possible to get around this problem by splitting the path before cloning it.

sasensi:
### Description

This change prevent `Maximum call stack exceeded` error by using a loop to add segment instead of passing segments array as argument.

#### Related issues

- Fixes #1493

### Checklist

### DIFF
--- a/src/path/Path.js
+++ b/src/path/Path.js
@@ -406,8 +406,9 @@ var Path = PathItem.extend(/** @lends Path# */{
         if (append) {
             // Append them all at the end.
             // Use a loop as it is the best way to handle big arrays (see #1493)
-            for (var i = 0; i < segs.length; i++)
+            for (var i = 0, l = segs.length; i < l; i++) {
                 segments.push(segs[i]);
+            }
         } else {
             // Insert somewhere else
             segments.splice.apply(segments, [index, 0].concat(segs));
--- a/src/path/Path.js
+++ b/src/path/Path.js
@@ -405,9 +405,13 @@ var Path = PathItem.extend(/** @lends Path# */{
         }
         if (append) {
             // Append them all at the end.
-            // Use a loop as it is the best way to handle big arrays (see #1493)
-            for (var i = 0, l = segs.length; i < l; i++) {
-                segments.push(segs[i]);
+            // Use a loop as the best way to handle big arrays (see #1493).
+            // Set future array length before the loop for better performances.
+            var originalLength = segments.length;
+            var offsetLength = segs.length;
+            segments.length += offsetLength;
+            for (var i = 0; i < offsetLength; i++) {
+                segments[originalLength + i] = segs[i];
             }
         } else {
             // Insert somewhere else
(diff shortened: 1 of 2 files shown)


src/path/Path.js
```diff
@@ -404,8 +404,10 @@ var Path = PathItem.extend(/** @lends Path# */{
                 this._updateSelection(segment, 0, segment._selection);
         }
         if (append) {
-            // Append them all at the end by using push
-            segments.push.apply(segments, segs);
+            // Append them all at the end.
+            // Use a loop as it is the best way to handle big arrays (see #1493)
+            for (var i = 0; i < segs.length; i++)
+                segments.push(segs[i]);
```
lehni: Some code-style remarks:

- Let's use blocks after `if` and `for` (unless the statement is super simple). I know the code-base doesn't do this currently in some places, but I tend to disagree with this by now. We shouldn't minify code by hand :) 

- Let's use this loop optimisation: `for (var i = 0, l = segs.length; i < l; i++)`

Result:

```js
for (var i = 0, l = segs.length; i < l; i++) {
    segments.push(segs[i]);
}
```

test/tests/Path.js
```diff
@@ -611,3 +611,13 @@ test('Path#arcTo(from, through, to); where from, through and to all share the sa
     }
     equals(error != null, true, 'We expect this arcTo() command to throw an error');
 });
+
+test('#1493 Path#add with 1000000 segments', function() {
+    var path = new Path();
+    for (var i=0; i<1000000; i++)
+    {
```
lehni: Could you bring the bracket to the same line as the `for`?  That's the convention in the paper.js code-base

src/path/Path.js
sasensi: - Sorry again for the style, I put that in my IDE rules too :)
- OK for the loop optimisation

test/tests/Path.js
sasensi: I added this one too.

sasensi:
Sure, I'd love to be part of the team, I have actually plenty of time to put into the project so I'll do my best to help.
Just, as this will be my first time as an open source project contributor, I might made small mistakes (as in those code style mistakes...) ;)

sasensi:
Ok, I'll do that for future commits.

PR #1545, opened 2018-10-06
Fix item selected position with matrix not applied

sasensi:
### Description

Make sure selected position is drawn in global coordinates system
whether item is in a group or not and whether matrix is applied or not.
Correct a bug introduced in 34679614c0db77387c72fb50e35daac6882e941e.

### Checklist

### DIFF
--- a/src/item/Item.js
+++ b/src/item/Item.js
@@ -4423,7 +4423,9 @@ new function() { // Injection scope for hit-test functions shared with project
             if (itemSelected)
                 this._drawSelected(ctx, mx, selectionItems);
             if (positionSelected) {
-                var point = mx._transformPoint(this.getPosition(true)),
+                // Get position in global coordinates system
+                var pos = this.getPosition(true),
+                    point = this.parent ? this.parent.localToGlobal(pos) : pos,
                     x = point.x,
                     y = point.y;
                 ctx.beginPath();


lehni:
Here an example of the bug fixed here: http://sketch.paperjs.org/#S/NU+xbsJADP0V6xaoGkXpGompQ6dKqB05BhMMRDnsk89AK8S/19c2Hiy/Zz/7+R4YzxT68DmRDafQhEH2FV9RQWkwWAHTDdZop/bDMfIx0fIeGTwOKuceNl0D3bb5o0yceOmc8jSTxVQmepUk2sNil3CYFpEfT5HriTZLGW0Ubgslx7T3o6YXihy5+jiqXPK/kbdaL6vM1b+NFnNO3+9oOn750AFTobk1b4bnVfUT2R/cKeGUZWQrod9sHz8=

lehni:
@sasensi I think this fix is more aligned with the rest of the code, and should also perform better: 91d3499191a312cd950651c7805b9290d9ab9ca6

Could you test if this works in your use-cases?

sasensi:
@lehni, yes it works now, I thinks it's due to the recent fixes we did on global matrix cache invalidation which was buggy at the time I did this PR.

lehni:
@sasensi my point was that my version of this fix and yours should behave the same, but I wasn't 100% certain. And it actually doesn't behave the same way, see this case:

```js
var rect = new Path.Rectangle({
    from: [0, 0],
    to: [100, 100],
    strokeColor: 'black',
    applyMatrix: false
})
rect.position.selected = true
rect.position += 100

var group = new Group(rect)
group.applyMatrix = false
group.position += 100
```

I'll revert back to your version, it was better :)

sasensi:
Oh, yes, you're right.
By the way, I should have posted a sketch with this PR, it would have been easier to track regressions.
Maybe we could add unit tests based on pixel comparisons for cases like this one ?

lehni:
@sasensi funny you mention this! I was just now thinking exactly about the pixel testing!

sasensi:
@lehni, a workaround for this problem could be to compare all pixels of a working case with the failing one ?

lehni:
@sasensi yes sounds great! 👍

PR #1546, opened 2018-10-06
Fix bounds error with nested empty items

### LINKED ISSUE #1467: Bounds break, when empty group is added
I've noticed strange behaviour of item's bounds, when its children list contains some empty groups.
I'm using JavaScript directly, tried only in Chrome.

Here is a [fiddle](https://jsfiddle.net/6g257vz4/51/) with a test case. Both layers in the fiddle should look the same.

sasensi:
### Description

- add an optional parameter to `Item#isEmpty()` allowing to check
emptiness recursively
- ignore all recursively empty items in bounds calculation

Bug is reproduced in this [Sketch](http://sketch.paperjs.org/#S/fZE/b8MgEMW/yokltmSRZHXVKUOXDpU6NhkIXGsazFkHthVF+e4F/1G6tBt39/i9x3ETXrUoavF+wagbUQlNJtfbLRhWI2jL2uGRj35QvFTwDB5HeFOxkYepU9w0+ohcT32yPha7CnZlBayM7UO9T+Wnde5AjriGDaPZ3MunzJ2Z8ky9N0EGdKgjmuQRucdJsWbxGPIE2y5e4Yup70IeZ8+XXBWPUzmzl9stDQhOXZEhEgw2yea8ed4xfSdLqXS0A75mlewo2GjJpxRZLWf1GiY0NC64OfVfmH/etKB+UzK3dwZalX5iXfXDIP3NmVFdurzeIOqP0/0H) where layer bounds should match circle bounds.

#### Related issues

- Closes #1467

### Checklist

### DIFF
--- a/src/item/Item.js
+++ b/src/item/Item.js
@@ -1043,7 +1043,9 @@ new function() { // Injection scope for various item event handlers
             options = options || {};
             for (var i = 0, l = items.length; i < l; i++) {
                 var item = items[i];
-                if (item._visible && !item.isEmpty()) {
+                // Item is handled if it is visible and not recursively empty.
+                // This avoid errors with nested empty groups (#1467).
+                if (item._visible && !item.isEmpty(true)) {
                     // Pass true for noInternal, since even when getting
                     // internal bounds for this item, we need to apply the
                     // matrices to its children.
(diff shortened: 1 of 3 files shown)


src/item/Item.js
```diff
@@ -2785,11 +2787,27 @@ new function() { // Injection scope for hit-test functions shared with project
...
+            return false;
+        // In recursive check, item is empty if all its children are empty.
+        } else {
+            for (var i = 0, l = children.length; i < l; i++) {
+                if (!children[i].isEmpty(true)) {
+                    return false;
+                }
+            }
+            return true;
+        }
```
sapics: @lehni This PR looks good!

By the way, I have a question about coding style.
This part can remove `else` like,
```javascript
if (!children || !children.length)
    return true;
if (!recursively)
    return false;
for (var i = 0, l = children.length; i < l; i++) {
    if (!children[i].isEmpty(true))
        return false;
}
return true;
```
Which is recommended style in paperjs?

lehni: I would write it this way:
```js
    isEmpty: function(recursively) {
        var children = this._children;
        var numChildren = children ? children.length : 0;
        if (recursively) {
            for (var i = 0; i < numChildren; i++) {
                if (!children[i].isEmpty(true)) {
                    return false;
                }
            }
        } else if (numChildren) {
            return false;
        }
        return true;
    },
```

But I'm not sure that overloading `isEmpty()` with an optional parameter is the way to go. Maybe we add another method for this check. I need to think about it a little.

sapics: I got it! Your function looks better.

By the way, when `retursively = true and numChildren = 0`, your function return true ;)
```js
    isEmpty: function(recursively) {
        var children = this._children;
        var numChildren = children ? children.length : 0;
        if (recursively) {
            for (var i = 0; i < numChildren; i++) {
                if (!children[i].isEmpty(true)) {
                    return false;
                }
            }
        }
        return !numChildren;
    },
```

lehni:
Wrapped up and merged in eeb26436b0805022ab20dc3a869f6fb2b7965670, closing

PR #1549, opened 2018-10-08
Fix mousedrag error with in front items

### LINKED ISSUE #1465: Mouse drag event problem.
This mouse drag event bug started from the release of paper.js v0.11.3:
    https://github.com/paperjs/paper.js/releases/tag/v0.11.3
The commit that caused this problem is:
   https://github.com/paperjs/paper.js/commit/fbd5eeb7ef12fe5f30565dd2b9fcdac3631c1e54
   Mouse Events: Fix item-based doubleclick events.
   Closes #1316

The bug is:
I have a canvas that I add svg graphics in it, canvas and svg graphics objects have the same event (mouse drag).
When the mousedrag event is on the canvas it moves the svg graphics, for the svg graphics objects is moving the svg graphic in the canvas.
At the svg graphics objects I have a point at the bottom right corner that when I click down and then mousedrag it resize the svg graphic object.
The problem starts when I do mouse down at the point of svg graphic obejct and then mouse drag if the mouse cursor leaves this point (because I move a little fast the mouse cursor) starts the mousedrag event of the canvas so for result i have two events runnin: the resize of the object and the canvas mousedrag that moves the objects in the canvas.

This bug started when updated the paper.js to the latest version. From v.0.11.3 and upper the bug started, below this version the mouse drag event does not have this problem.

sasensi:
When dragging item A behind item B, mousedrag event is emitted on
item B instead of mousemove.
Demonstration in this [Sketch](http://sketch.paperjs.org/#S/vZJNa8MwDIb/isilLctCL4ORssPWXQeDDXZoe3AcpfHiWMVW0kHpf5+c0A/YaTvMYIxfS9b7yD4kTrWY5Mlbg6zrJE00lXHfKw+GsX2EB3C4h1fFdbY0XlucHtZ+7UCGRsfoc+gN7rNxAzewur2fpzDfpKcwr0rThRzu5mepMtYuyZIkT8grt8XJ+SywpwY/TMm15Jxlci/UBXz2aptD1TnNhtwUeyk7gwPXJmQ7Ciaq4nnQRTCOF8d4xXG2iEucJ7anP7D9Dq2w3X+CDTDx5B2/+AqFxL1j8RNrwNhv0AMwUC9g0af4QGCCgLLUCGWM3XkqLLbZBWKo/KMt0hV59UtfPrvApjJaRddSeIycXPmVr1Z4VM1wX0jy1eb4DQ==).

Bug introduced in fbd5eeb7ef12fe5f30565dd2b9fcdac3631c1e54
Closes #1465

### DIFF
--- a/src/view/View.js
+++ b/src/view/View.js
@@ -1271,11 +1271,12 @@ new function() { // Injection scope for event handling on the browser
                     point, prevPoint)
             // Next handle the hit-item, if it's different from the drag-item
             // and not a descendant of it (in which case it would already have
-            // received an event in the call above).
+            // received an event in the call above). Translate mousedrag to
+            // mousemove, since drag is handled above.
             || hitItem && hitItem !== dragItem
                 && !hitItem.isDescendant(dragItem)
-                && emitMouseEvent(hitItem, null, type, event, point, prevPoint,
-                    dragItem)
+                && emitMouseEvent(hitItem, null, type === 'mousedrag' ?
+                'mousemove' : type, event, point, prevPoint, dragItem)
             // Lastly handle the mouse events on the view, if we're still here.
             || emitMouseEvent(view, dragItem || hitItem || view, type, event,
                     point, prevPoint));
--- a/src/view/View.js
+++ b/src/view/View.js
@@ -1276,7 +1276,7 @@ new function() { // Injection scope for event handling on the browser
             || hitItem && hitItem !== dragItem
                 && !hitItem.isDescendant(dragItem)
                 && emitMouseEvent(hitItem, null, type === 'mousedrag' ?
-                'mousemove' : type, event, point, prevPoint, dragItem)
+                    'mousemove' : type, event, point, prevPoint, dragItem)
             // Lastly handle the mouse events on the view, if we're still here.
             || emitMouseEvent(view, dragItem || hitItem || view, type, event,
                     point, prevPoint));
--- a/test/tests/Interactions.js
+++ b/test/tests/Interactions.js
@@ -334,3 +334,14 @@ test('Item#onMouseDrag() is not triggered after mouse up', function(assert) {
     expect(1);
 });
 
+test('Item#onMouseDrag() is not triggered if mouse down was on another item', function(assert) {
+    var item = new Path.Rectangle(new Point(0, 0), new Size(10));
+    item.fillColor = 'red';
+    var item2 = item.clone().translate(10);
+    item2.onMouseDrag = function(event) {
+        throw 'this should not be called';
+    };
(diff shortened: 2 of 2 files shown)


src/view/View.js
```diff
@@ -1271,11 +1271,12 @@ new function() { // Injection scope for event handling on the browser
...
             // and not a descendant of it (in which case it would already have
-            // received an event in the call above).
+            // received an event in the call above). Translate mousedrag to
+            // mousemove, since drag is handled above.
             || hitItem && hitItem !== dragItem
                 && !hitItem.isDescendant(dragItem)
-                && emitMouseEvent(hitItem, null, type, event, point, prevPoint,
-                    dragItem)
+                && emitMouseEvent(hitItem, null, type === 'mousedrag' ?
+                'mousemove' : type, event, point, prevPoint, dragItem)
```
lehni: @sasensi this line should be indented by 4 more spaces, as it's inside a method call.

lehni:
@sasensi I'm generally quite worried about changing event code, since we don't have mouse based unit tests in place to test for common scenarios, thus we wouldn't detect breaking changes there.

sasensi:
@lehni, about adding unit tests for mouse events, I 100% agree with you.
So I was starting adding them and I encountered a first problem when trying to trigger events: we have class name collisions with native events. this is problematic when working in "loaded" context (after running `gulp load`) because the global scope is polluted with all `Paper.js` classes.
What we can't do in this context is instantiating native events and dispatching them:
```javascript
var event = new Event();
element.dispatchEvent(event);
```
I see 3 not perfect solutions:
- not running tests in loaded context (but this would degrade development experience)
- rely on old [document.createEvent()](https://developer.mozilla.org/fr/docs/Web/API/Document/createEvent) method to instantiate native events (but this is deprecated and can be removed from browser API)
- rename `Paper.js` event classes to avoid names collisions (but this can break backward compatibility)

lehni:
@sasensi that's unfortunate about `Event`. I don't think people use the class though, they just use the name in the reference to look up what properties it offers when it is received in callbacks. We could rather easily rename it, but what should it be called? `PaperEvent`? `BaseEvent`? hmmm...

Another option is to patch `load.js` to store a reference to the native `Event` object, e.g. under `NativeEvent`, and then we could use that in the tests?

lehni:
@sasensi sounds like a good plan, thanks!

sasensi:
@lehni, here is the PR #1570 adding tests.

src/view/View.js
lehni: @sasensi could you address this?

sasensi: Ok, done :)

lehni:
🎉

PR #1551, opened 2018-10-08
Fix double click behavior is different from native

### LINKED ISSUE #1451: The doubleclick event should behave like the native DOM dblclick event
I looked over the Paper codebase and find out that `doubleclick` event is emitted if the previously clicked item is the same as the current clicked item and **the time difference between the first `mouseup` event and the second `mousedown` event** is less than 300 ms. This leads to an unexpected result when you can press the mouse button, hold it for a couple of minutes, drag it over the clicked item for whatever distance, then release the mouse button, immediately click on the previous item one more time and `doubleclick` will still be emitted. That's not how the native DOM `dblclick` event works (see expected result).

Here is a simple test-case where you can easily test both events:
[https://jsfiddle.net/kcyrLbty/6/](https://jsfiddle.net/kcyrLbty/6/)

# Expected result
I did some tests and it looks like browsers compare **the time difference between the first `mousedown` event and the second `mouseup` event** (which must be less than 300 ms). Also, they take into consideration the distance between the first mouse down point and the second mouse up point (which must be less than a few pixels).

If you think this issue is important and needs to be fixed, let me know please. I have already fixed it locally and can open a PR.

sasensi:
### Description

Changes based on @DSnopov suggestion in #1451.
- time delta is calculated between 2 down events (instead of between up and down)
- a check is made to only emit double click if distance between 2 down points is under 3 pixels

#### Related issues

- Closes #1451

### Checklist


src/view/View.js
```diff
@@ -1391,12 +1391,7 @@ new function() { // Injection scope for event handling on the browser
             if (mouse.down && inView || mouse.up && downPoint) {
                 emitMouseEvents(this, hitItem, type, event, point, downPoint);
                 if (mouse.down) {
-                    // See if we're clicking again on the same item, within the
-                    // double-click time. Firefox uses 300ms as the max time
-                    // difference:
-                    dblClick = hitItem === clickItem
-                        && (Date.now() - clickTime < 300);
-                    downItem = clickItem = hitItem;
```
lehni: @sasensi this removes the check for clicks on the same item entirely... Are you sure we want this? Is it fine to do so because double-clicks are only considered within 3 pixels distance? But what if the first click was on a different item?

sasensi: No, you're right, I hadn't noticed that. But even if I don't manage to make it fail practically, from a theoretical point of view, it would be better to also check that item is the same for both clicks.
In new version, I also ensure that triple clicking an item don't produce 2 double click events.

lehni:
@sasensi this looks good now! Could you add unit tests for it also, now that we have the infrastructure for them in place?

sasensi:
@lehni I added unit test, ignore failing build as it is only in node v11 (see explanation [here](https://github.com/orgs/paperjs/teams/contributors/discussions/8)).

lehni:
@sasensi could you rebase this on top of `develop`, so the tests pass?

RickMohr:
Can this get merged?

PR #1552, opened 2018-10-08
Fix children global matrices are not cleared

### LINKED ISSUE #1448: Item#localToGlobal() returns same results before and after its parent transforms when applyMatrix = false
1. create rectangle path and add it to group
2. get global point at the bottom-right of the rectangle
3. change the position of the group
4. get global point at the bottom-right of the rectangle. It return the same position as (2).

# Expected result
I think it should return different position because its parent changed the position.
Here is my test code.

```
    // 1. Create path and add it to group
    let path = Path.Rectangle({
      position: new Point(200,200),
      size: new Size(100,100),
      fillColor: 'black'
    })
    let group = new Group({
      children: [path]
    })

    // 2. Get global point at the bottom-right of the rectangle
    let point1 = path.localToGlobal(path.getPointAt(0))
    console.log(point1)

    // 3. Move the group
    group.position = new Point(300,200)

    // 4. Get global point at the bottom-right of the rectangle again,
    //    But it returns the same result as (2).
    let point2 = group.children[0].localToGlobal(path.getPointAt(0))
    console.log(point2)

```

sasensi:
### Description

Between a parent transformation and next view update, children global matrix was invalid, leading to wrong global coordinates calculation.
Here is the [Sketch](http://sketch.paperjs.org/#S/pVNNb9swDP0rhC9xgMx1su3iYYchh5wKFG1vbQ+yTNtaFNGQmLRd0f8+Ska+ih6CLYADUaTee/x6y5zaYFZld2tk3WezTFMT7asrmBew9KgYYVDcg3INqKYBw8AEnaftEB79o9spL3e4gZ9wI3HFLWpWrrOYv0U3yG+gYNiQq8DhM9yQcZzPy3IG8jed7aOC+YNjxJ2cPglojbVLsuQrmNRW6fUket6nP/YykijRETFW8XyUoHtjG48i4SGKfTqAqmGwr9eKvXmpoFU24Alo/KQSiwJWyNBZqpWVbEQ/KClDj1ATM22+eNP1DNSmO7+vwF5XejEXYZG6sKSVvadVQsvTVYecqvKL83I6UmtygSxKdJeP788kfS3gmnaY+I69SKeCvXLBSuPGGpZnD7/9Yy6gOmXc7CylxX+mtDgqS4MRx2dpvD6dHY2OUTo+1uDQNq8asw2VDMin4+GxmRwcNCht+LWCsvj+obmX0S4upa3tFi/njdD3+MInrCSsjgXpFhtoiNPWRdRkhJ62VmyxTNuijDMf2ZLSCnYGn4tR/cH1exvYtEarcQkno/t0e2Txa9n1dQIJWfXw9P4X).

#### Related issues

- Closes #1448

### Checklist

### DIFF
--- a/src/item/Item.js
+++ b/src/item/Item.js
@@ -216,8 +216,8 @@ new function() { // Injection scope for various item event handlers
         if (flags & /*#=*/ChangeFlag.GEOMETRY) {
             // Clear cached bounds, position and decomposed matrix whenever
             // geometry changes.
-            this._bounds = this._position = this._decomposed =
-                    this._globalMatrix = undefined;
+            this._bounds = this._position = this._decomposed = undefined;
+            this._clearGlobalMatrix();
         }
         if (cacheParent
                 && (flags & /*#=*/(ChangeFlag.GEOMETRY | ChangeFlag.STROKE))) {
@@ -972,6 +972,20 @@ new function() { // Injection scope for various item event handlers
                     : this._parent || this._symbol && this._symbol._item,
             mx = parent ? parent.getViewMatrix().invert() : matrix;
         return mx && mx._shiftless();
+    },
+
+    /**
+     * Clears cached global matrix for item and all its descendants.
+     * This is used to avoid invalid coodrinates calculation (#1448).
+     */
+    _clearGlobalMatrix: function() {
+        this._globalMatrix = undefined;
+        var children = this.children;
+        if (children) {
+            for (var i = 0, l = children.length; i < l; i++) {
+                children[i]._clearGlobalMatrix();
+            }
+        }
     },
 
     statics: /** @lends Item */{
--- a/test/tests/Item.js
+++ b/test/tests/Item.js
@@ -937,3 +937,11 @@ test('Item#position pivot point and caching (#1503)', function() {
     item.translate(5, 5);
     equals(item.position, new Point(5, 5));
 });
+
+test('Children global matrices are cleared after parent transformation', function() {
+    var item = Path.Rectangle(new Point(0, 0), new Size(100));
+    var group = new Group({ children: [item], applyMatrix: false });
+    equals(item.localToGlobal(item.getPointAt(0)), new Point(0, 100));
+    group.translate(100, 0);
+    equals(item.localToGlobal(item.getPointAt(0)), new Point(100, 100));
+});


src/item/Item.js
```diff
@@ -216,8 +216,8 @@ new function() { // Injection scope for various item event handlers
         if (flags & /*#=*/ChangeFlag.GEOMETRY) {
             // Clear cached bounds, position and decomposed matrix whenever
             // geometry changes.
-            this._bounds = this._position = this._decomposed =
-                    this._globalMatrix = undefined;
+            this._bounds = this._position = this._decomposed = undefined;
+            Item._clearGlobalMatrixCache(this);
```
lehni: Could this be an instance method instead, and maybe have a shorter name? 

`this._clearGlobalMatrix();`

src/item/Item.js
```diff
@@ -1029,6 +1029,19 @@ new function() { // Injection scope for various item event handlers
...
+         * Clears cached global matrix for item and all its descendants.
+         * This is used to avoid invalid coodrinates calculation (#1448).
+         */
+        _clearGlobalMatrixCache: function(item) {
+            item._globalMatrix = undefined;
+            if (item.children && item.children.length > 0) {
+                for (var i = 0, l = item.children.length; i < l; i++) {
+                    Item._clearGlobalMatrixCache(item.children[i]);
+                }
+            }
```
lehni: Here you repeatedly access `item.children`. For performance and also code minification it is better to reference it in a local variable first:
`var children = item.children;`
Also, you don't need to check for `length` if  your next statement is a loop that only loops ifthe length is > 0, so:
```js
            var children = this.children;
            if (children) {
                for (var i = 0, l = children.length; i < l; i++) {
                    children[i]._clearGlobalMatrix();
                }
            }
   
 ```

sasensi:
@lehni, I made the code changes and this time, I replaced the original commit instead of adding a new one.
Is that the good way to do go ?

lehni:
@sasensi yes that's fine!

lehni:
I would still like to think about this a little, to see if there's not a way to avoid having to loop through children in `_changed()` events. This probably has a massive impact on performance, otherwise.

lehni:
@sasensi I am going to merge this, but I have found a way to remove the need to call `_clearGlobalMatrix()` in every `_changed()` event that I will follow up with.

lehni:
@sasensi check the changes in c5b822da799db5c89d840a88476708b91a664d51, I think this should perform better

Also, I want to tackle this new TODO next:

https://github.com/paperjs/paper.js/blob/c5b822da799db5c89d840a88476708b91a664d51/src/item/Item.js#L220-L221

PR #1562, opened 2018-10-11
Fix item global matrix error

### LINKED ISSUE #1561: Set opacity of a symbol of a group of an empty path will cause all bounds calculation affected by the symbol's transforms.
When I make a symbol reference to a **group with an empty path** as it's children, if I do not change the opacity of the symbol, everything works fine.
Once I changed the opacity of the symbol, all transform applied to symbol will directly affect the bounds of other items. Here is the code to reproduce the problem:

```
let v1 = new paper.Path()
let g1 = new paper.Group({ applyMatrix: false, children: [v1] })
let s1 = new paper.SymbolItem(g1)
s1.opacity = 0.5
s1.rotate(30)
s1.skew(10)

let v2 = new paper.Path.Circle({ center: [0, 0], radius: 100, fillColor: 'red' })
v2.selected = true
```

error

PS:
1. If the opacity not change, the result is OK.
2. If the symbol directly reference to the path, the result is OK.
3. If the path is non-empty, the result is OK.

sasensi:
### Description
Bug happen when item is drawn after an empty symbol that should be drawn in a separate canvas context (partial opacity or special blend mode).
As bounds are empty, symbol drawing process is interrupted but its global matrix is not removed from the stack.

#### Related issues

- Closes #1561

### Checklist

### DIFF
--- a/src/item/Item.js
+++ b/src/item/Item.js
@@ -4298,8 +4298,12 @@ new function() { // Injection scope for hit-test functions shared with project
             // Apply the parent's global matrix to the calculation of correct
             // bounds.
             var bounds = this.getStrokeBounds(viewMatrix);
-            if (!bounds.width || !bounds.height)
+            if (!bounds.width || !bounds.height) {
+                // Item won't be drawn so its global matrix need to be removed
+                // from the stack (#1561).
+                matrices.pop();
                 return;
+            }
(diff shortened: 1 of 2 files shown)


lehni:
Look good, thanks!

PR #1568, opened 2018-10-15
Remove PaperScript code from core version

sapics:
### Description

This PR is a code minification for core version of paper.js.
It is also better to avoid error.

### DIFF
--- a/src/core/PaperScope.js
+++ b/src/core/PaperScope.js
@@ -203,9 +203,11 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
      * @param {Object} [option] the compilation options
      */
     execute: function(code, options) {
-        var exports = paper.PaperScript.execute(code, this, options);
-        View.updateFocus();
-        return exports;
+/*#*/   if (__options.paperScript) {
+            var exports = paper.PaperScript.execute(code, this, options);
+            View.updateFocus();
+            return exports;
+/*#*/   }
     },
 
     /**


src/core/PaperScope.js
```diff
@@ -203,9 +203,11 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
...
      */
     execute: function(code, options) {
-        var exports = paper.PaperScript.execute(code, this, options);
-        View.updateFocus();
-        return exports;
+/*#*/   if (__options.paperScript) {
+            var exports = paper.PaperScript.execute(code, this, options);
+            View.updateFocus();
+            return exports;
+/*#*/   }
```
sasensi: @lehni and @sapics, I use this change opportunity to ask you what is the purpose of these marks: /*#*/
I guess it has something to do with the building system but I wasn't able to find how.

lehni: @sasensi these are instructions for the preprocessor. The code that follows is not executed in the browser but at preprocessing time, when creating the final file that contains the code from all the included files. The `include()` statements are also executed in this way, and so are the constant replacements, look for `/*#=*/`.

When we started working on paper.js, none of the now popular packages existed (webpack, rollup, etc). I came up with prepro.js as a way to be able to directly work in the browser during development without any preprocessing steps needed through `load.js` / `prepro.js`, and then have a way to build a single clean file from all the source code files.

sasensi: @lehni, crystal clear explanation, thanks ! 👍

PR #1571, opened 2018-10-16
Fix group selected bounds and position color

sasensi:
### Bug
Group selected color was applied differently to its bounds and position depending on whether it had children or not when selected color was set.
This resulted in an unpredictable behaviour from a user point of view.
Here is the [sketch](http://sketch.paperjs.org/#S/jZLLTsMwEEV/ZZRNUhHSh9RNEBu6YMMCCXZtF649adwGu7InLajqvzNuHgoPISJlYfvOvcfjOUdGvGGURy97JFlGaSStCuvxGKRDQQgz2DpbH/zKrcxRuGY1hXsweILHsEhGd+GwOcg2tjbKZx4rlISKheRqHCoO1mvS1vyq6SNmvNsWyMoabFPCz3CF0JU2W5DCY7sllAJZ6kpBoZ0noBINeCToY6StrBuAcMUiFCThKs+CymyhnawwWU5SmKxTmE9GX+7WOS2CEfPFDlU8xDpZt/+G9ZNgyNdD9ymzP1Jawf+5u9Y4J8wW25cEb+GEzBjag6B0UaBDI9EPMogrfMUDkLDndDZv/A7O7pgtE5L0EZ/EB7r+OZn0qPGUSTSEro+/Mlpt6BXfKTmHLeBPWlYZyiF+sFR2ZM3wgDAKeltf2poftRRHDD0DzxPbNDJOO7dDCMiH+XADoRu30zk3pJPtak+60FIEY45upHE4vozuePo3PPP7q5mP8uX68gk=) reproducing the bug.

screenshot-127 0 0 1-8080-2018 10 16-15-54-10

### Change description
- When `item.setSelectedColor()` is called, value is now always stored in `item._style._values`, independently from the fact that item has children or not.
- An helper method `compareCanvas()` is added to the test suite to allow comparing selection rendering of a known working case to a failing one.
Two provided callbacks are executed in a dedicated `<canvas>`/`Project` context and both results are compared with `resemble.js`.

### Checklist


test/helpers.js
```diff
@@ -283,6 +306,36 @@ var compareItem = function(actual, expected, message, options, properties) {
...
+ * Run each callback in a separated canvas context and compare both outputs.
+ * This can be used to do selection drawing tests as it is not possible with
+ * comparePixels() method which relies on the item.rasterize() method which
+ * ignores selection.
+ * @param width the width of the canvas
+ * @param height the height of the canvas
+ * @param callback1 the function producing the expected result
+ * @param callback2 the function producing the actual result
+ */
+var compareCanvas = function(width, height, callback1, callback2) {
```
lehni: @sasensi I think it would be better to call them `expectedCallback` and `actualCallback` then?

test/helpers.js
```diff
@@ -283,6 +306,36 @@ var compareItem = function(actual, expected, message, options, properties) {
...
+        var project = new Project(canvas);
+        callback();
+        project.view.update();
+        var imageData = canvas.getContext('2d').getImageData(0, 0, width, height);
+        CanvasProvider.release(canvas);
+        project.remove();
+        return imageData;
+    }
+
+    compareImageDatas(
```
lehni: @sasensi the plural of `data` is ...  `data` :) Let's call it `compareImageData()`. It's clear that you need to provide two since you compare them.

sasensi: Ok :D

sasensi:
@lehni, wait before merging this one because it has reference to `CanvasProvider` in test file. That would make the build crash again...

lehni:
👍

sasensi:
@lehni, I fixed it so it can now be merged.

lehni:
@sasensi why not simple using the already existing `comparePixels()` method which takes two items?

PR #1573, opened 2018-10-17
Add documentation for PaperScope#setup(size)

### LINKED ISSUE #1412: PaperScope#setup(element|string) documentation doesn't mention how to use the library inside a web-worker
Commit [8fb7c41](https://github.com/paperjs/paper.js/commit/8fb7c415379f6d0711cb8ec9842cd8529a42c51b) introduced a change that made it possible to create a paper.js project w/o a HTMLCanvasElement (i.e. so the library can be used with web-workers, etc.). This is discussed further in #634.

However, the documentation has not been updated to indicate that this is possible. I think the [PaperScope#setup(element)](http://paperjs.org/reference/paperscope/#setup-element) documentation should be updated (i.e. now it is possible to pass in a `Size` object instead of a canvas element/string).

sasensi:
### Description
Documentation did not mention that a size can be passed as argument to PaperScope#setup() and Project#initialize() for usage of paper.js in web workers.

#### Related issues

- Closes #1412

### Checklist

### DIFF
--- a/src/core/PaperScope.js
+++ b/src/core/PaperScope.js
@@ -249,9 +249,10 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
      * Sets up an empty project for us. If a canvas is provided, it also creates
      * a {@link View} for it, both linked to this scope.
      *
-     * @param {HTMLCanvasElement|String} element the HTML canvas element this
-     * scope should be associated with, or an ID string by which to find the
-     * element.
+     * @param {HTMLCanvasElement|String|Size} element the HTML canvas element
+     * this scope should be associated with, or an ID string by which to find
+     * the element, or the size of the canvas to be created for usage in a web
+     * worker.
      */
     setup: function(element) {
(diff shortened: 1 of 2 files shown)


src/core/PaperScope.js
```diff
@@ -249,9 +249,9 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
      * Sets up an empty project for us. If a canvas is provided, it also creates
      * a {@link View} for it, both linked to this scope.
      *
-     * @param {HTMLCanvasElement|String} element the HTML canvas element this
+     * @param {HTMLCanvasElement|String|Size} element the HTML canvas element this
```
lehni: @sasensi pretty sure this is over 80chars... does the hinter not complain?

sasensi: That's right, sorry.
About the hinter, if you talk about `gulp jshint` task, I think it has no rule about that.

PR #1570, opened 2018-10-15
Add mouse interaction tests

sasensi:
### Description
- Tests common mouse interactions scenarios to prevent regressions when making changes. These tests are not run in node context.
- Prevent name collision between Javascript native classes and Paper.js classes (Event and MouseEvent) by patching load.js.
- Uses a polyfill for MouseEvent which is missing in PhantomJS.
- Adds View._clearState() method and use it in tests to make sure that each new test is started with a fresh state.

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the JSHint rules (`npm run jshint` passes)

PR #1574, opened 2018-10-17
Fix mouse interactions tests error

sasensi:
### Description
Mouse interactions tests only passed in `gulp load` context.
This make sure that native event classes are used in built context.
This also remove the reference to `CanvasProvider` which is not available in built context.

#### Related pull request

- Fixes bug introduced in #1570

### Checklist

### DIFF
--- a/test/tests/load.js
+++ b/test/tests/load.js
@@ -64,4 +64,6 @@
 /*#*/ include('Numerical.js');
 
 // There is no need to test interactions in node context.
-/*#*/ if (!isNode) include('Interactions.js');
+if (!isNode) {
+    /*#*/ include('Interactions.js');
+}
--- a/test/tests/load.js
+++ b/test/tests/load.js
@@ -64,4 +64,4 @@
 /*#*/ include('Numerical.js');
 
 // There is no need to test interactions in node context.
-if (!isNode) /*#*/ include('Interactions.js');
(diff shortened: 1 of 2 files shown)


test/tests/load.js
```diff
@@ -64,4 +64,4 @@
 /*#*/ include('Numerical.js');
 
 // There is no need to test interactions in node context.
-if (!isNode) /*#*/ include('Interactions.js');
+/*#*/ if (!isNode) include('Interactions.js');
```
lehni: @sasensi does this work? I don't think `isNode` is a pre-process-time global variable.

Could you try write this code like this, on three lines?

```js
if (!isNode) {
    /*#*/ include('Interactions.js');
}
```

sasensi: I don't understand, test now work on my local travis but still fail on online one !
I'll try your idea.

lehni: @sapics also pointed out weirdness with Travis... I really have no clue what's going on.

lehni: I think `/*#*/` needs to be on its own line, not nested with other code that is evaluated at execution time.

sasensi: I'll try to revert changes in a separated branch to see if failure is really related to code changes.

lehni:
@sasensi I've merged it. Let's see how this goes :)

PR #1578, opened 2018-10-20
Minor code minifier

sapics:
### Description
We can minify some codes which relate matrix.decompose(),
because matrix.decompose() must return non-null object

### DIFF
--- a/src/basic/Matrix.js
+++ b/src/basic/Matrix.js
@@ -672,12 +672,11 @@ var Matrix = Base.extend(/** @lends Matrix# */{
     },
 
     /**
-     * Attempts to decompose the affine transformation described by this matrix
-     * into `scaling`, `rotation` and `skewing`, and returns an object with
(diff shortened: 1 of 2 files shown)


src/basic/Matrix.js
```diff
@@ -672,12 +672,11 @@ var Matrix = Base.extend(/** @lends Matrix# */{
     },
 
     /**
-     * Attempts to decompose the affine transformation described by this matrix
-     * into `scaling`, `rotation` and `skewing`, and returns an object with
-     * these properties if it succeeded, `null` otherwise.
+     * Decompose the affine transformation described by this matrix into
```
lehni: @sapics minor spelling mistake: `Decomposes`. After that it's good to go! Thanks for spotting this!

sapics: Thanks for noticing my mistake in grammar!

PR #1579, opened 2018-10-20
Fix transform attribution in SvgExport

sapics:
### Description
There are some cases that matrix cannot be described by affine transformation.
To avoid such cases, we should not use affine transformation created by
matrix.decompose(), but use matrix transformation.

### Checklist

### DIFF
--- a/src/svg/SvgExport.js
+++ b/src/svg/SvgExport.js
@@ -36,29 +36,7 @@ new function() {
             trans = null;
         }
         if (!matrix.isIdentity()) {
-            // See if we can decompose the matrix and can formulate it as a
-            // simple translate/scale/rotate command sequence.
-            var decomposed = matrix.decompose();
-            if (decomposed) {
-                var parts = [],
-                    angle = decomposed.rotation,
-                    scale = decomposed.scaling,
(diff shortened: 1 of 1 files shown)


lehni:
@sapics I don't agree with this change... We should try decomposition but if it does not work, we should use the matrix. And if the decomposition is wrong, then we should find those cases and fix them.

Do you have a test-case where this goes wrong currently? We should also create a unit test for that.

sapics:
`matrix.decompose()` returns always object. Thus, it would be necessary.
Anyway, I will try to create test-case!

lehni:
@sapics yes I am aware of that now, but maybe we should change that to only return if the transformation is valid, as the rest of the code was assuming.

sapics:
@lehni I have not success to create a test-case yet.
Now, I get the other error in SvgExport.js line 33, during creating a test-case.
`matrix._inverseTransform(trans);` returns null, when matrix is singular.

sapics:
I created this PR, just I found in src code. I should create an error-case first, sorry.

Now, I create an issue in #1580 which causes error.
When I or someone fix the error, new PR would be created.
Thus, I close this PR.

PR #1586, opened 2018-10-24
Travis: Remove node v9 and add node v10

sapics:
### Description
From the [release schedule of nodejs](https://github.com/nodejs/Release#release-schedule), I remove nodejs v9 because of end-of-life.
We add v10 to travis, because `stable` version means v11. (v11 is released in 23th, Oct. !)


lehni:
👍

PR #1593, opened 2018-11-05
Fix Item#importJSON override Item#insert

### LINKED ISSUE #1392: `importJSON` function overwrites `insert` method in `path` with a variable.

Create an empty path, check whether `path.insert` is a function (as it should be). Now, import a path using `path.importJSON()`. `path.insert` should still be a function, however, it's now a boolean variable assigned the value false. This bug breaks in particular the possibility of running `path.divideAt()` on the imported path, or any other function that would require `path.insert()`.

# Expected result

After path.importJSON(), the function path.insert should still exist, however, it's replaced by a boolean variable.

# Additional information

The culprit is this part in paperjs's code:
```
	importJSON: function(json, target) {
		return Base.deserialize(
				typeof json === 'string' ? JSON.parse(json) : json,
				function(ctor, args, isRoot) {
					var useTarget = isRoot && target
							&& target.constructor === ctor,
						obj = useTarget ? target
							: Base.create(ctor.prototype);
					if (args.length === 1 && obj instanceof Item
							&& (useTarget || !(obj instanceof Layer))) {
						var arg = args[0];
						if (Base.isPlainObject(arg))
							arg.insert = false;
					}
					(useTarget ? obj.set : ctor).apply(obj, args);
					if (useTarget)
						target = null;
					return obj;
				});
	},
```

`arg.insert` is overwriting the `path.insert` function. Replacing `arg.insert = false` by `arg.insertFlag = false`, solves the issue of the overwritten `path.insert` function. In particular, it is now possible to run `path.divideAt()`.

Tested on Mac OS Sierra 10.12.6 (16G29), Chrome 57.0.2987.98 (64-bit).

sasensi:
### Description
When calling `item.importJSON()`, `item.insert()` was overriden along with item properties. An exclude 
 object is passed to `Base.set()` to prevent that from happening.

#### Related issues

- Closes #1392

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## Prebuilt version
+
+### Fixed
+
+- Prevent `Item#importJSON()` from overriding `Item#insert()` (#1392).
+
+### Added
+
 ## `0.11.8`
 
 ### News
--- a/src/core/Base.js
+++ b/src/core/Base.js
@@ -558,8 +558,17 @@ statics: /** @lends Base */{
                     if (args.length === 1 && obj instanceof Item
                             && (useTarget || !(obj instanceof Layer))) {
                         var arg = args[0];
-                        if (Base.isPlainObject(arg))
+                        if (Base.isPlainObject(arg)) {
                             arg.insert = false;
+                            // When using target, make sure `item.insert()`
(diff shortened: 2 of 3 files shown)


lehni:
Wrapped up and merged in 0eae0b6e4ddd024e40600abff115d0d2154ec3cd, closing

PR #1612, opened 2018-11-21
Add typescript definition generation

### LINKED ISSUE #985: Add TypeScript type definitions
Paper.js is a pretty popular module. It'd be fantastic if it had typings on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), letting people use it with frameworks like Angular 2 with greater ease.

sasensi:
### Description

This add a gulp task (`gulp docs:typescript`) to automatically generate
a typescript definition for the library. This should solve the problem
of having an out of sync type definition when we change the API.

This task takes advantage of existing JSDoc parsing to generate a
temporary file which is later formatted through a mustache template to
generate the final definition. This definition is then tested by
compiling a typescript file that use it.
The generated definition is added to the `gulp zip` task in order to be
published along with the bundled library.

So 2 new dev-dependencies are added with this change: `mustache` and
`typescript` packages. Using node and mustache to generate the
definition instead of relying on existing templating system is
motivated by a better development experience, with easier debugging
possibilities... through the usage of more modern tools.

As a side note, support of "rest parameters" (when a parameter can be
present multiple times) is added to existing JSDoc parser in order to
support this pattern on typescript side (E.g. for `Color#set()` method
which accept any sequence of parameters that is supported by `Color`
constructors).

#### Related issues

- Depends on https://github.com/paperjs/jsdoc/pull/4 merge
- Closes #985

### Checklist

## Example

[Here](https://gist.github.com/sasensi/bec4d11423362e0f660a134ba1aad0de) is a demonstration of the generated type definition.

### DIFF
--- a/gulp/tasks/dist.js
+++ b/gulp/tasks/dist.js
@@ -22,6 +22,7 @@ gulp.task('zip', ['clean:zip', 'dist'], function() {
             gulp.src([
                 'dist/paper-full*.js',
                 'dist/paper-core*.js',
+                'dist/index.d.ts',
                 'dist/node/**/*',
                 'LICENSE.txt',
                 'examples/**/*',
--- a/gulp/tasks/dist.js
+++ b/gulp/tasks/dist.js
@@ -22,7 +22,7 @@ gulp.task('zip', ['clean:zip', 'dist'], function() {
             gulp.src([
                 'dist/paper-full*.js',
                 'dist/paper-core*.js',
-                'dist/index.d.ts',
+                'dist/paper.d.ts',
                 'dist/node/**/*',
                 'LICENSE.txt',
                 'examples/**/*',
--- a/gulp/typescript/typescript-definition-generator.js
+++ b/gulp/typescript/typescript-definition-generator.js
@@ -1,0 +1,317 @@
+/**
+ * This script generates a type definition by taking JSDoc roughly parsed data,
+ * formatting it and passing it to a mustache template.
+ */
+
+const fs = require('fs');
+const mustache = require('mustache');
+
+// Retrieve JSDoc data.
+const data = JSON.parse(fs.readFileSync(__dirname + '/typescript-definition-data.json', 'utf8'));
+const classes = data.classes;
+let globals = data.global.properties;
+
+// Format classes.
+classes.forEach(cls => {
+    // Format class.
+    // Store name as `className` and not simply `name`, to avoid name conflict
+    // in static constructors block.
+    cls.className = cls._name;
+    // Store closest parent if there is one.
+    cls.extends = cls.inheritsFrom && cls.inheritsFrom.length > 0
+        ? cls.inheritsFrom[0]
+        : null;
+    // Store comment using class tag as description.
+    cls.comment = formatComment(cls.comment, 'class');
+
+    // Build a filter for deprecated or inherited methods or properties.
+    const filter = _ => !_.deprecated && _.memberOf == cls.alias && !_.isNamespace;
+
+    // Format properties.
+    cls.properties = cls.properties
+        .filter(filter)
+        .map(_ => ({
+            name: _._name,
+            type: formatType(_.type),
+            static: formatStatic(_.isStatic),
+            readOnly: formatReadOnly(_.readOnly),
+            comment: formatComment(_.comment)
+        }));
+
+    // Format methods.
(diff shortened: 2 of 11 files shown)


sasensi:
The type definition skeleton was adapted from the existing definition in https://github.com/DefinitelyTyped/DefinitelyTyped repo (module declaration, export, ...) in order to preserve backward compatibility.

sasensi:
Also note that this PR doesn't take care (yet) of the link between `Paper.js` node package and the type definition.

sapics:
@sasensi Thanks for great work! It would help a lot for typescript user!

By the way, I have a question about type naming rule.
In typescript, we need to use lowercase for [basic types](https://www.typescriptlang.org/docs/handbook/basic-types.html) as {number}, {boolean}.
On the other hand, we use upper-camel as {Number}, {Boolean}, in paper.js jsdoc.
Should we change the upper-camel to lowercase for basic types in jsdoc?

lehni:
@sasensi this is super cool! I will merge this now, shall we take care of publishing it to DefinitelyTyped next?

gulp/tasks/dist.js
```diff
@@ -22,6 +22,7 @@ gulp.task('zip', ['clean:zip', 'dist'], function() {
             gulp.src([
                 'dist/paper-full*.js',
                 'dist/paper-core*.js',
+                'dist/index.d.ts',
```
lehni: @sasensi shouldn't this be called `paper.d.ts`?

gulp/typescript/typescript-definition-generator.js
```diff
@@ -0,0 +1,317 @@
...
+    cls.className = cls._name;
+    // Store closest parent if there is one.
+    cls.extends = cls.inheritsFrom && cls.inheritsFrom.length > 0
+        ? cls.inheritsFrom[0]
+        : null;
+    // Store comment using class tag as description.
+    cls.comment = formatComment(cls.comment, 'class');
+
+    // Build a filter for deprecated or inherited methods or properties.
+    const filter = _ => !_.deprecated && _.memberOf == cls.alias && !_.isNamespace;
```
lehni: @sasensi I think this use of `_` for these method arguments is rather confusing. I first thought you're using the underscore library :) 

Could you replace it with `it`? I've seen this being used elsewhere, it's short, and it's verbally correct, as you're iterating over an array and each entry is `it`. Sounds good?

sasensi:
@lehni, when I did this feature, I was thinking about removing DefinitelyTyped entry for `Paper.js` and publishing a definition along with the releases instead. The problem with DefinitelyTyped is that it doesn't handle library version history (see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/7719).
But maybe we should have both ?

gulp/tasks/dist.js
sasensi: @lehni, I named it `index.d.ts` thinking that it would allow automatic binding when publishing along with node module.
See explanation [here](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html):

> Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so.

But we can name it differently and add a `types` entry in `package.json` file if you think this is better ?

lehni:
@sasensi I don't know what's better, since I don't use it personally :) Your suggestion sounds easier to maintain since we have full control over it?

gulp/tasks/dist.js
lehni: Well, it doesn't live at the root here, does it? It looks like we can use the `"types"` field in `package.json` to indicate this:

https://medium.com/cameron-nokes/the-30-second-guide-to-publishing-a-typescript-package-to-npm-89d93ff7bccd#a8f1

Then I'd go with `"types": "dist/paper.d.ts"`

sasensi: Yes, you're right :)

gulp/typescript/typescript-definition-generator.js
sasensi: Ok, I pushed the change.

lehni: Thanks! I just released `v0.12.0`, kind of a shame I didn't include this in it as well... Will  merge now!

PR #1620, opened 2018-11-29
Implement tweening

arnoson:
as discussed IRL


lehni:
Thanks @arnoson! Let's merge this in and take it from there. I think the docs need a bit more work, but I can also join the effort :)

PR #1622, opened 2018-11-30
Add Tween documentation

sasensi:
### Description

Add documentation for the new `Tween` class.
I didn't documented `Tween#handleFrame()` and `Tween#update()` because it seemed to me that these didn't have public use cases for the moment.

#### Related issues

- Depends on https://github.com/paperjs/jsdoc/pull/5 merge

### Checklist

### DIFF
--- a/src/item/Tween.js
+++ b/src/item/Tween.js
@@ -13,7 +13,8 @@
 /**
  * @name Tween
  *
- * @class Allows to tween properties of an Item between two values
+ * @class Allows tweening {@link Item} properties between two states for a given
+ * duration. Tween instance is returned by {@link Item#tween(from,to,options)}.
  */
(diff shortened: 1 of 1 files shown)


src/item/Tween.js
```diff
@@ -214,6 +256,40 @@ var Tween = Base.extend(Emitter, /** @lends Tween# */{
         return this;
     },
 
+    /**
+     * {@grouptitle Event Handling}
+     *
+     * Attaches an event handler to the tween.
+     *
+     * @name Tween#on
```
arnoson: @lehni suggested that we document this as Tween#onUpdate

sasensi: Ok, I made the change.

sasensi: About not documenting `Tween#update()` until we remove the need to call start, I agree with you.

Issue #1628, opened 2018-12-11
Shadow is clipped if item has stroke and fill and opacity < 1


georeith:
Example with fill and stroke:

http://sketch.paperjs.org/#V/0.11.8/S/bY9RC4IwEMe/ytiLBSJWTy16qQ8Q5KP6MOeZw7XJNpMSv3vT3IPQcRzc78fB/wYs6RMwwUkDltU4xEyV0/6iGmlglsqHAHRGEnqU1LSF6O7pZsgkcmX4BwhKd3EcIjfy8IdbxaX9wysuxFUJpZ1zZuno4L1qKeP2TRw7LshYrRpYjgINZeBFTUvVe1GIDtbmVlUGphD7OUO+khfRuStnMjluT+71QgNt5tgGkzQfvw==

Example with just fill:

http://sketch.paperjs.org/#V/0.11.8/S/bY/BCoMwDEB/pfTiBiJuO61jl+0DBvOoHmqNs9i1UutkE/990elBWAiBvEdI0lPNn0AZjSpwoqQ+FSYf+xe3xIJwXD8UkDPR0JGo5DUE94Vu+kQTjEZ+gJF4F4Y+wZL6P1wbqd0fXkilrkYZiw7NnMFh8abmQro3Q3acUVPy3HTzkJepFryVuRVFA+Ou/bQqXcmLanEKTaKH7Qk/zCzwarquoSxOhy8=

PR #1631, opened 2018-12-12
Fix shadows being clipped for items with fill, stroke and opacity < 1

georeith:
### Description

Fixes an issue where a shadow would be clipped if the item has a stroke and a fill and opacity < 1.

#### Related issues

- Relates to https://github.com/paperjs/paper.js/issues/1628

### Checklist


src/util/Numerical.js
```diff
@@ -159,7 +159,13 @@ var Numerical = new function() {
...
          */
         KAPPA: 4 * (sqrt(2) - 1) / 3,
-
+        /**
+         * Amount to multiply radius of blur by to account for error
+         * in gaussian blur kernel filter approximation
+         *
+         * https://dbaron.org/log/20110225-blur-radius
+         */
+        GAUSSIAN_BLUR_COEFFICIENT: 1.4,
```
lehni: Could you explain how you ended up with `1.4` from the cited document? I don't see any mention of 1.4 in there?

src/item/Item.js
```diff
@@ -4347,6 +4347,16 @@ new function() { // Injection scope for hit-test functions shared with project
             // Apply the parent's global matrix to the calculation of correct
             // bounds.
             var bounds = this.getStrokeBounds(viewMatrix);
+            // Account for drop shadow
+            var style = this.getStyle();
+            if (style.hasShadow()) {
+                var shadowRadius = style.getShadowBlur() * Numerical.GAUSSIAN_BLUR_COEFFICIENT;
```
lehni: We don't really need to put this into `Numerical.GAUSSIAN_BLUR_COEFFICIENT` since it's only used in one place.

Could you inline the value directly here, and add the comment about the reason right above it?

src/util/Numerical.js
lehni: Could it actually be `Math.SQRT2 ( === Math.sqrt(2))`?

src/item/Item.js
lehni: @georeith still waiting on feedback from you here

francoly:
@lehni I have also encountered this problem in paper 0.12.3. When I used both stroke and transparency, the sharp edges of the shapes were truncated。
image

src/util/Numerical.js
georeith: @lehni apologies I seem to have missed this.

1.4 was just in my testing what I found to be sufficiently large to cover extreme cases. 

I don't know how to calculate the actual error but this number doesn't need to precise, it's a tradeoff between precision and speed where the larger number is more accomadating but slower as it creates a larger in memory canvas.

src/item/Item.js
georeith: @lehni sorry I forgot to come back to this just seen it pop up in my feed again. I can make this change.

georeith:
@francoly that is the same issue but not fixed by this as I only expand for blur. Your issue is that bounds don't take into account stroke mitering which allows the corners of strokes to extend further than the stroke width away from a path.

src/item/Item.js
lehni: That'd be great, thank you!

lehni:
@francoly this is a separate issue that is linked to stroke style somehow. Could you open a separate issue?

PR #1650, opened 2019-04-25
Fixing the this.setItem bug

supreetsingh247:
### Description
I am using Paper JS in my production project. I keep getting this error -

Uncaught (in promise) TypeError: this.setItem is not a function

Screenshot 2019-04-25 at 6 25 35 PM

This will fix the issue.

#### Related issues

- Relates to <link_to_referenced_issue>

### Checklist

### DIFF
--- a/src/item/SymbolDefinition.js
+++ b/src/item/SymbolDefinition.js
@@ -60,7 +60,9 @@ var SymbolDefinition = Base.extend(/** @lends SymbolDefinition# */{
         this._id = UID.get();
         this.project = paper.project;
         if (item)
-            this.setItem(item, dontCenter);
+            try {
+                this.setItem(item, dontCenter);
+            } catch(e) {}
     },
 
     _serialize: function(options, dictionary) {


supreetsingh247:
Can someone merge this PR. This is really creating a problem for us in production. Thanks.

lehni:
Could you provide a test case that reproduces the bug? simply swallowing the error with no trace is not the solution here, and this MR will not be merged in this state.

supreetsingh247:
I will let you know the error case.

supreetsingh247:
I have found the issue. The paperJS context, if installed in the 'window' scope defines 'Symbol' function in the global/window context. I am also using another library which has a similar definition(again installed globally). PaperJs is installed after that dependency, hence when that particular library tries to call the Symbol function, it hits the 'paper-full.js' function instead of its own function. Hence the error.
I think using such common names in the global context is a bit risky. I am not sure how to fix this issue. I just want to understand if anything can be done in paperJS level to prevent any such future conflicts. If not, I think this MR can be closed.

supreetsingh247:
I have setup a new global variable and passed it as scope to paper.install. That solves the issue. Thanks for the help @lehni

lehni:
great, thanks @supreetsingh247 👍

PR #1651, opened 2019-04-25
Item.contains returns false when matrix not invertible instead of blowing up

danfstucky:
### Description
Currently, if an irregular shape is present, it can cause the following error with hover/click events:
`TypeError: Cannot read property 'isInside' of null`
It occurs because the shape is not invertible so the `_inverseTransform` fn returns null which causes the error downstream.
paper.js should return false instead of failing.

#### Related issues
Relates to https://github.com/paperjs/paper.js/pull/1156 (I believe this old PR was trying to address the same issue)

### Checklist

### DIFF
--- a/src/item/Item.js
+++ b/src/item/Item.js
@@ -1803,8 +1803,12 @@ new function() { // Injection scope for various item event handlers
      */
     contains: function(/* point */) {
         // See CompoundPath#_contains() for the reason for !!
-        return !!this._contains(
-                this._matrix._inverseTransform(Point.read(arguments)));
+        if (this._matrix.isInvertible()) {
+            return !!this._contains(
+                    this._matrix._inverseTransform(Point.read(arguments)));
+        } else {
+            return false;
+        }
     },
 
     _contains: function(point) {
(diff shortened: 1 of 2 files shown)


src/item/Item.js
```diff
@@ -1803,8 +1803,12 @@ new function() { // Injection scope for various item event handlers
...
     contains: function(/* point */) {
         // See CompoundPath#_contains() for the reason for !!
-        return !!this._contains(
-                this._matrix._inverseTransform(Point.read(arguments)));
+        if (this._matrix.isInvertible()) {
+            return !!this._contains(
+                    this._matrix._inverseTransform(Point.read(arguments)));
+        } else {
+            return false;
+        }
```
lehni: Could you shorten this to:

```js
var matrix = this._matrix;
return (
	matrix.isInvertible() &&
	!!this._contains(this._matrix._inverseTransform(Point.read(arguments)))
);
```

lehni: (the first line is so that minification works more efficiently, because local variables can be minified, while properties can't)

danfstucky: Made the change. Didn't know that about the minification too. Thanks.

lehni: Thanks!

PR #1658, opened 2019-06-05
Fix node tests crash with node latest versions.

sasensi:
### Description
`canvas` package v1 was no longer compatible with node v11 & v12.
As it was a dependency of both `jsdom` and `resemblejs`, this commit updates these packages to their latest versions which work with `canvas` v2.
As these updates include some API breaking changes, this commit also updates `paper` code accordingly.

### Checklist

### DIFF
--- a/src/core/PaperScope.js
+++ b/src/core/PaperScope.js
@@ -110,7 +110,6 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
             // `agent.node` is expected to be true in node context.
             if (agent.jsdom) {
                 agent.node = true;
-                delete agent.jsdom;
             }
         }
     },
--- a/src/core/PaperScope.js
+++ b/src/core/PaperScope.js
@@ -87,7 +87,7 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
             // here: { chrome: true, webkit: false }, Mozilla missing is the
             // only difference to jQuery.browser
             user.replace(
-                /(opera|chrome|safari|webkit|firefox|msie|trident|atom|node)\/?\s*([.\d]+)(?:.*version\/([.\d]+))?(?:.*rv\:v?([.\d]+))?/g,
+                /(opera|chrome|safari|webkit|firefox|msie|trident|atom|node|jsdom)\/?\s*([.\d]+)(?:.*version\/([.\d]+))?(?:.*rv\:v?([.\d]+))?/g,
                 function(match, n, v1, v2, rv) {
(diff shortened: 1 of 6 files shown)


sasensi:
@lehni, this should fix our prebuilt branch workflow.
I tested with node v11 locally and hope it will pass in Travis with other node versions.
That said, it seems that the Travis build was not triggered by the pull request, could it be because this commit includes a change to the `travis.yml` file ?

sasensi:
Forget about Travis build not being triggered, there was a typo in `travis.yml` file.

lehni:
Looks like we worked on exactly the same things at the same time!

src/core/PaperScope.js
```diff
@@ -105,6 +105,13 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
                 delete agent.webkit;
             if (agent.atom)
                 delete agent.chrome;
+            // In node context, JSDOM user agent no longer include "node" word,
+            // it is now called "jsdom". So we need to replace it because
+            // `agent.node` is expected to be true in node context.
+            if (agent.jsdom) {
+                agent.node = true;
+                delete agent.jsdom;
```
lehni: @sasensi could you remove the `delete`? we can keep this.

src/item/Raster.js
```diff
@@ -408,6 +408,8 @@ var Raster = Item.extend(/** @lends Raster# */{
             crossOrigin = this._crossOrigin;
         if (crossOrigin)
             image.crossOrigin = crossOrigin;
+        // We need to avoid setting image source to null, otherwise an unwanted
+        // error would be thrown in node context.
```
lehni: perhaps better to mention that this is a JSDOM bug?

src/core/PaperScope.js
lehni: Even simpler and shorter:
`agent.node = !!agent.jsdom`

src/item/Raster.js
sasensi: How would you rephrase that as a good english speaker ? :)

src/core/PaperScope.js
sasensi: ok

src/item/Raster.js
lehni:
```
// Prevent setting image source to `null`, as this isn't supported by browsers,
// and it would actually throw exceptions in JSDOM.
//TODO: Look into fixing this exception in JSDOM

PR #1637, opened 2019-01-08
Fix SVG imported gradients default values

sasensi:
### Description

Add default values based on SVG specification document.

#### Related issues

- Closes #1632

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the JSHint rules (`npm run jshint` passes)

PR #1661, opened 2019-06-06
Fix SVG linear gradient import x2 default value.

### LINKED ISSUE #1660: SVG linear gradient import fails when x2 is not set
SVG linear gradient import doesn't follow [SVG spec](https://www.w3.org/TR/SVG11/pservers.html#LinearGradients).
`x2` default value should be `100%`.
Originally reported [here](https://stackoverflow.com/questions/56476565/import-svg-with-gradient-stroke-to-paper-js-project).

https://jsfiddle.net/u1bcew0t/

screenshot-jsfiddle net-2019 06 06-14-35-33

# Expected result
screenshot-127 0 0 1-8080-2019 06 06-14-36-35

sasensi:
### Description
As for SVG spec, linear gradient `x2` default value should be `100%`.
This commit also include a rename of the internal variable `isNode`, used in test, to `isNodeContext`.
This was needed because the new version of `resemblejs` exposes a `isNode` function which was conflicting with it.

#### Related issues
- Closes #1660 

### Checklist

### DIFF
--- a/src/svg/SvgImport.js
+++ b/src/svg/SvgImport.js
@@ -174,7 +174,14 @@ new function() {
             highlight = getPoint(node, 'fx', 'fy', true, scaleToBounds);
         } else {
             origin = getPoint(node, 'x1', 'y1', false, scaleToBounds);
-            destination = getPoint(node, 'x2', 'y2', false, scaleToBounds);
+            // As SVG spec states, x2 default value should be `100%`:
+            // https://www.w3.org/TR/SVG11/pservers.html#LinearGradients
+            // so we manually parse each attribute instead of using getPoint().
+            var x2 = node.hasAttribute('x2')
+                ? getValue(node, 'x2', false, false, scaleToBounds)
+                : rootSize.width;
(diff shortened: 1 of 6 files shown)


lehni:
@sasensi this collides in multiple ways with #1637, which I would like to merge first:

- Both introduce the file `gradients-3.svg`
- #1637 adds `defaultX` / `defaultY` to `getPoint()`, which I think is the better approach here too. Do you agree?

So let's merge  #1637 first, and then rewrite this PR here once that's done, OK?

lehni:
I've merged this in 5a3cf624aa75c4e68b4d1659b26a587d01065091 now... It looks like #1637 actually fully fixed this issue already, and nothing new needed to be done :) but the test is now also included. thanks!

sasensi:
@lehni, sorry for the late answer.
Indeed, this issue seemed familiar to me when I encountered it but I forgot that I did the other PR some time ago.
It is good that you take the time to clear all those PRs 👍

PR #1669, opened 2019-06-14
Fix typescript definition issues

### LINKED ISSUE #1659: Some missing types of build-in typescript definition file.
Since version `0.12.1` and upcoming major/minor release there are build-in typings for TypeScript. #985 

Heres my post from this issue:

It seems to work generally. The typed-file is detected and works for import. But a few definitions are missing. Here just a few examples from my project. 

```ts
import * as paper from 'paper';
```

```ts
// Missing setup-method in global scope of paper namespace. also install-method.
paper.setup(canvas);
// Quick'n'Dirty Workaround
(paper as any).setup(canvas);
```
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5344bfc80508c53a23dae37b860fb0c905ff7b24/types/paper/index.d.ts#L66

It works with `setup()` but maybe I should use it in another way? But `setup()` exists. 

```ts
// Argument of Path is only a "object" not a interface with defined properties.
// It's ok, no problem, but a weak type. e.g. no autocompletion / suggestion in vs code :(
// Note: Not all properties are required. So they should be marked as optional typescript: `?`
const path = new paper.Path({
  strokeColor: '#ffff00',
  strokeWidth: 5,
  strokeCap: round
});

// Throws an error: Argument of type `number[]` is not assignable to parameter of type `Point | Segment`.
// But this is not true. It is possible to assign number[].
path.add([100, 200]);
// Quick'n'Dirty Workaround :(
(path as any).add([100, 200]);
```
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5344bfc80508c53a23dae37b860fb0c905ff7b24/types/paper/index.d.ts#L2418

That is interesting. The `types/paper` definition file also has only `Segment | Point`.
But for sure `number[]` also works. I've tested it. The `types/paper` throws no error. 

Reference: 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5344bfc80508c53a23dae37b860fb0c905ff7b24/types/paper/index.d.ts

The latest (paper documentation)[http://paperjs.org/reference/global/] also has no setup-method. Don't know if the last documentation had it. I think it's just a missing declaration in the docs / comment.
Update: Wait... there it is: http://paperjs.org/reference/paperscope/ 
But setup() also exists in global scope. 🤔 Is this maybe a predefined paperscope?
It seems so, yes. This means `setup` and `install` is missing in the types.

Some versions. (package.json)
```json
{
  "paper": "0.12.1",
  "typescript": "3.2.4"
}
```
VS-Code Info incl. OS, Node, etc.
```text
Version: 1.35.0 (user setup)
Commit: 553cfb2c2205db5f15f3ee8395bbd5cf066d357d
Date: 2019-06-04T01:17:12.481Z
Electron: 3.1.8
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
OS: Windows_NT x64 10.0.17763
```

Any questions? I'm glad if I can help.

### LINKED ISSUE #1663: [typescript] Missing declaration of class constructible via PaperScope

Write TypeScript code that constructs objects using a `PaperScope` instead as documented [here](https://github.com/paperjs/paper.js/blob/develop/dist/paper.d.ts#L3225)

https://github.com/dobesv/paperjs-ts-repro

Check repro1.ts

# Expected result

Should be able to construct objects via the `PaperScope` object as documented.  Even though there is a type error, the code works as expected.

# Additional information

I had to make the same fix to the DefinitelyTyped files recently:

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35361

You can refer to that for an example of how the types of these classes can be written.

### LINKED ISSUE #1664: [typescript] Raster.onLoad is declared non-nullable
# Steps to reproduce

Set a raster's `onLoad` handler to `null` in typescript code, with `"strictNullChecks": true` set in `tsconfig.json`.

https://github.com/dobesv/paperjs-ts-repro/blob/master/raster-onload-null.ts

# Expected result

Assigning `onLoad` to `null` type checks without errors, as this is a valid operation which removes the `onLoad` handler from the `Raster`.

# Additional information

`"strictNullChecks": true` is a fairly popular setting in typescript to help detect possible runtime errors involving `null` or `undefined` values.

However, with this setting enabled, you cannot set the `onLoad` handler to `null` on `Raster` without getting a typescript error.

### LINKED ISSUE #1667: [typescript] event.modifiers has object type

Try to inspect `ToolEvent.modifiers` within typescript code.

https://github.com/dobesv/paperjs-ts-repro/blob/master/toolevent-modifiers.ts

# Expected result

You should be able to check which modifiers were pressed at the time of a `ToolEvent` when writing TypeScript code, without casting to `any` or getting type errors.

# Additional information

Currently `ToolEvent.modifiers` is declared as the type `object`, which in typescript has no properties accessible.  It can only be used by casting to some other type, such as `any` or some custom interface created to workaround this issue.

The definitions in `@types/paper` have an interface in place for this purpose, perhaps you can follow that example.

Or, at the very least, declare it as `any` rather than object.  At least there won't be type errors (but there might be more runtime errors).

sasensi:
The first step of TypeScript definition improvements.

Depends on https://github.com/paperjs/jsdoc/pull/7 merge.

Closes #1667
Closes #1664
Closes #1663
Closes #1659

### DIFF
--- a/dist/paper-full.js
+++ b/dist/paper-full.js
@@ -1,1 +1,17081 @@
-../src/load.js
+/*!
+ * Paper.js v0.12.2-fix/typescript-definition-issues - The Swiss Army Knife of Vector Graphics Scripting.
+ * http://paperjs.org/
+ *
+ * Copyright (c) 2011 - 2019, Juerg Lehni & Jonathan Puckey
+ * http://scratchdisk.com/ & https://puckey.studio/
+ *
+ * Distributed under the MIT license. See LICENSE file for details.
+ *
+ * All rights reserved.
+ *
+ * Date: Fri Jun 14 15:03:35 2019 +0200
+ *
+ ***
+ *
+ * Straps.js - Class inheritance library with support for bean-style accessors
+ *
+ * Copyright (c) 2006 - 2019 Juerg Lehni
+ * http://scratchdisk.com/
+ *
+ * Distributed under the MIT license.
+ *
+ ***
+ *
+ * Acorn.js
+ * https://marijnhaverbeke.nl/acorn/
+ *
+ * Acorn is a tiny, fast JavaScript parser written in JavaScript,
+ * created by Marijn Haverbeke and released under an MIT license.
+ *
+ */
+
+var paper = function(self, undefined) {
+
+self = self || require('./node/self.js');
+var window = self.window,
+	document = self.document;
+
+var Base = new function() {
+	var hidden = /^(statics|enumerable|beans|preserve)$/,
+		array = [],
+		slice = array.slice,
+		create = Object.create,
+		describe = Object.getOwnPropertyDescriptor,
+		define = Object.defineProperty,
+
+		forEach = array.forEach || function(iter, bind) {
+			for (var i = 0, l = this.length; i < l; i++) {
+				iter.call(bind, this[i], i, this);
+			}
+		},
+
+		forIn = function(iter, bind) {
+			for (var i in this) {
+				if (this.hasOwnProperty(i))
+					iter.call(bind, this[i], i, this);
+			}
+		},
+
+		set = Object.assign || function(dst) {
+			for (var i = 1, l = arguments.length; i < l; i++) {
+				var src = arguments[i];
+				for (var key in src) {
+					if (src.hasOwnProperty(key))
+						dst[key] = src[key];
+				}
+			}
+			return dst;
+		},
+
+		each = function(obj, iter, bind) {
+			if (obj) {
+				var desc = describe(obj, 'length');
+				(desc && typeof desc.value === 'number' ? forEach : forIn)
+					.call(obj, iter, bind = bind || obj);
+			}
+			return bind;
+		};
+
+	function inject(dest, src, enumerable, beans, preserve) {
+		var beansNames = {};
+
+		function field(name, val) {
+			val = val || (val = describe(src, name))
+					&& (val.get ? val : val.value);
+			if (typeof val === 'string' && val[0] === '#')
+				val = dest[val.substring(1)] || val;
+			var isFunc = typeof val === 'function',
+				res = val,
+				prev = preserve || isFunc && !val.base
+						? (val && val.get ? name in dest : dest[name])
+						: null,
+				bean;
+			if (!preserve || !prev) {
+				if (isFunc && prev)
+					val.base = prev;
+				if (isFunc && beans !== false
+						&& (bean = name.match(/^([gs]et|is)(([A-Z])(.*))$/)))
+					beansNames[bean[3].toLowerCase() + bean[4]] = bean[2];
+				if (!res || isFunc || !res.get || typeof res.get !== 'function'
+						|| !Base.isPlainObject(res)) {
+					res = { value: res, writable: true };
+				}
+				if ((describe(dest, name)
+						|| { configurable: true }).configurable) {
+					res.configurable = true;
+					res.enumerable = enumerable != null ? enumerable : !bean;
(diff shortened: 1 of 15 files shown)


lehni:
Thanks @sasensi! I am looking into this now. But this strikes me as a little strange:

```js
strokeScaling: boolean | null
```
I don't think booleans should specify `| null`? They're always either `true` or `false`...

lehni:
These `null` settings are also wrong:

```js
    class Point  {
        /** 
         * The x coordinate of the point
         */
        x: number | null

        /** 
         * The y coordinate of the point
         */
        y: number | null
```

How to handle this?

lehni:
@sasensi I'm going to merge this now because I already updated JSDoc and it crashes on `develop` currently, but let's keep discussing these cases and maybe address it in a separate PR?

lehni:
Oh damn, this PR included built dist files. This should never be the case... I'll have to manually revert it, but please exclude builds from commits.

sasensi:
@lehni, about strange nullable cases like the one you mention above, yes I am aware of that, as I stated here: https://github.com/paperjs/paper.js/issues/1664#issuecomment-501627907 maybe we should add extra information in our doc for nullable/not nullable properties. And maybe the library documentation could benefit from that as well. What do you think ?  
The current solution is only made to prevent the typescript compilator from complaining in strict null mode but it can definitely be improved.

About the unwanted files in the PR, oops sorry, I was not careful enough.
I will avoid that for future PRs.

lehni:
I like the idea of adding support fur `nullable` information.

Could we back-port the JSDoc way of doing so perhaps? I found some mention here:

https://github.com/phetsims/tasks/issues/935

lehni:
The docs are now here: https://devdocs.io/jsdoc/tags-type

image

Maybe it's enough to only support the first version?

sasensi:
@lehni, ok, great.
Yes, doing it the JSDoc way makes total sense as it will be compatible with new JSDoc version when we will migrate to it.
I agree with you that specifying nullable types is enough (`?type`), and we will assume that everything else is not nullable.
Do you want me to handle adding this to our jsdoc ?

lehni:
@sasensi yes, that would be great, if you feel like doing so?

sasensi:
Yes, totally, maybe before starting, I can ask you if you have a method for debugging in the Java + Rhino context because I am currently doing it the old `console.log` way which is far from the more productive :)

lehni:
Sadly that's the only way I've been doing so also :(

sasensi:
Ok, no problem, it was just in case 😁 
Maybe the part on which I will need your help is to decide which data needs to be marked as nullable in the doc.
And I think that we can do the same kind of review for what needs to be marked as `readonly`, for example, I don't think that the user should manually set [paperScope.project](http://paperjs.org/reference/paperscope/#project) or [paperScope.tool](http://paperjs.org/reference/paperscope/#tool), should he ?

lehni:
One thing you can do to speed development up is run the docs generation on an alternative folder that only contains one file in which you test all kinds of tags and formats. I've done so in the past, and it works a lot faster.

lehni:
Readonly should be generated automatically, for anything that has only a getter and no setter. Does this not work? For those things that are purely properties in the code, we should indeed provide it.

sasensi:
Yes, readonly generation for properties that has only a getter works but I think that there are missing readonly tags for other properties.

sasensi:
@lehni, here is the PR: https://github.com/paperjs/paper.js/pull/1677

PR #1677, opened 2019-06-26
Add support for nullable in documentation

sasensi:
### Description
The goal was mainly to improve the generated typescript definition by only declaring a property nullable when it really is.
This adds support for `?Type` as nullable type syntax in JSDoc parser.
This also declares some nullable properties in the doc.
I only found cases where it made sense for `Color` and `Function` types but there might be others that I missed.

#### Related issues

- Depends on https://github.com/paperjs/jsdoc/pull/8 merge
- Relates to https://github.com/paperjs/paper.js/issues/1664

### Checklist

### DIFF
--- a/gulp/typescript/typescript-definition-generator.js
+++ b/gulp/typescript/typescript-definition-generator.js
@@ -154,13 +154,41 @@ function parseType(type, options) {
     // `...` prefix and add `[]` as a suffix:
     // - `...Type` => `Type[]`
     // - `...(TypeA|TypeB)` => `(TypeA|TypeB)[]`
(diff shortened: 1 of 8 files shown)


Domvel:
@sasensi Thanks 👍 I hope for a release as soon as possible. 🙂 Otherwise I going insane. 🤪
Ok, I disabled the strict mode to avoid it. But it would be great if this fix will be merged and released.

lehni:
I've just merged this. I need to update the JSDoc sub-module in an additional commit, then we're good to go. Thank you @sasensi!

PR #1678, opened 2019-06-30
fix typescript definition for Point.isEmpty

bboyle:
#### Related issues

- Relates to <link_to_referenced_issue>

### Checklist

- [ ] Code conforms with the JSHint rules (`npm run jshint` passes)

### DIFF
--- a/dist/paper.d.ts
+++ b/dist/paper.d.ts
@@ -2323,7 +2323,7 @@ declare module paper {
          * 
          * @return Boolean
          */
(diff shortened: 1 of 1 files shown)


sasensi:
Hi, thank you for your report.
Unfortunately, your way to solve the problem was not the proper one as the typescript definition is directly generated from code documentation.
So the problem was a syntax error in method documentation.
I close this PR because it will be fixed with https://github.com/paperjs/paper.js/pull/1679.

PR #1689, opened 2019-07-11
Add View#fitBounds()

### LINKED ISSUE #1688: View fit bounds of Item.
Is there a way to fit the view to a drawing?
There is a method `fitBounds` of `Item`. This resizes (scales) the drawing.
But I want a method of `View` to fit the drawing without resizing it. 

```js
// e.g.
paper.project.activeLayer.fitBounds(paper.view.bounds.scale(0.9));
// or
group.fitBounds(view.bounds);
// etc.
```

[Demo](https://stackblitz.com/edit/js-paper-fit-view-issue?embed=1&file=index.js)

Demo instructions:
Use the GUI on the top right to draw a path. (e.g. a square)
Hit the "Fit to View" button and draw again. 

I do not want to scale the drawing. I want to add a new point (0, -100) relative to the last point without calculating the scaled coordinate. I could get the last point and add 100. But I have to scale 100 to the actual scale factor of the drawing. It's not the best method to do this.

I want to fit the view to the drawing. Not the drawing to the view. 
So I suggest a new feature. The method `fitBounds` for the View. 
I do not know how that should work. Zoom, center. 
Maybe I'm thinking in the wrong direction too.

I just want to fit the view to the drawing while drawing a path. By adding absolute points. This does not work if the drawing is always resized. (See demo)
The solution above (`item.fitBounds(view.bounds)`) works if I call it after complete the drawing, but not while the drawing is in progress. Because it's resized.

demo

Paper version `0.12.3`

sasensi:
#### Related issues

- Closes #1688

### Checklist

### DIFF
--- a/src/view/View.js
+++ b/src/view/View.js
@@ -680,6 +680,38 @@ var View = Base.extend(Emitter, /** @lends View# */{
     },
 
     /**
+     * Transform the view so that its {@link #bounds} fit within the
+     * specified rectangle.
+     *
+     * @param {Rectangle} rectangle
+     * @param {Boolean} [fill=false]
+     *
+     * @example {@paperscript height=100}
+     * // Fitting the view to the bounding rectangle of an item's bounding:
+     *
+     * // Create a circle at the top left corner of the view.
+     * var item = new Path.Circle({
+     *    center: [0, 0],
+     *    radius: 50,
+     *    fillColor: 'orange'
+     * });
+     *
+     * // Make view fit its bounds.
+     * view.fitBounds(item.bounds);
+     */
+    fitBounds: function(rectangle, fill) {
+        rectangle = Rectangle.read(arguments);
+        var bounds = this.getBounds();
+        var horizontalRatio = bounds.width / rectangle.width;
+        var verticalRatio = bounds.height / rectangle.height;
+        var scale = fill
+            ? Math.max(horizontalRatio, verticalRatio)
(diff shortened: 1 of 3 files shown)


src/view/View.js
```diff
@@ -679,6 +679,38 @@ var View = Base.extend(Emitter, /** @lends View# */{
...
+    fitBounds: function(rectangle, fill) {
+        rectangle = Rectangle.read(arguments);
+        var bounds = this.getBounds();
+        var horizontalRatio = bounds.width / rectangle.width;
+        var verticalRatio = bounds.height / rectangle.height;
+        var scale = fill
+            ? Math.max(horizontalRatio, verticalRatio)
+            : Math.min(horizontalRatio, verticalRatio);
+        this.translate(bounds.center.subtract(rectangle.center));
+        this.scale(scale);
```
lehni: @sasensi this is the same code as on Item, yes? Can't we reuse it and safe some space, e.g. through a function on `Rectangle`?

lehni:
@sasensi I added one comment, otherwise this looks good to go, thanks!

src/view/View.js
sasensi: @lehni, actually, I wrote a different code than `item.fitBounds()` because `View` didn't have a `setBounds()` method and the item version is relying on this method.
So if you want to reuse the code, we also need to find a way to share the `item.setBounds()` code with the `View` class.
Maybe making `view.bounds` settable makes sense too ?

PR #1713, opened 2019-09-27
feat: allows import core version in typescript

anion155:
### Description

Allows typescript users to import core version of paper as `import paper from 'paper/dist/paper-core';`

### Checklist

- [ ] Code conforms with the JSHint rules (`npm run jshint` passes)

### DIFF
--- a/dist/paper-core.d.ts
+++ b/dist/paper-core.d.ts
@@ -1,2 +1,1 @@
-import paper from 'paper';
-export default paper;
+import 'paper';
--- a/dist/paper-core.d.ts
+++ b/dist/paper-core.d.ts
@@ -1,0 +1,2 @@
+import paper from 'paper';
+export default paper;
--- a/gulp/tasks/dist.js
+++ b/gulp/tasks/dist.js
@@ -23,6 +23,7 @@ gulp.task('zip', ['clean:zip', 'dist'], function() {
                 'dist/paper-full*.js',
                 'dist/paper-core*.js',
                 'dist/paper.d.ts',
+                'dist/paper-core.d.ts',
                 'dist/node/**/*',
                 'LICENSE.txt',
                 'examples/**/*',
--- a/gulp/typescript/typescript-definition-template.mustache
+++ b/gulp/typescript/typescript-definition-template.mustache
@@ -68,3 +68,6 @@ declare module paper {
 declare module 'paper' {
     export = paper
 }
+declare module 'paper/dist/paper-core' {
+    export = paper
+}


anion155:
@sasensi

sasensi:
Hi, thank you for your contribution.
Could you elaborate a bit more about what this change does (how it works, in what case that would be usefull...) ?
And could you imagine a test that would validate this usage ?
I already use a TypeScript file compilation as a [test](https://github.com/paperjs/paper.js/blob/develop/gulp/typescript/typescript-definition-test.ts) so maybe this could be based on the same principle ?

anion155:
Paper by default export paper-full which is has acorn, PaperScript, and all this beautiful staff, which is not always really needed (see [paperjs/paper.js#which-version-to-use](https://github.com/paperjs/paper.js#which-version-to-use)). I needed a way to import core version of the lib, but I did loose typings.

This file will allow to import paper-core with typings as default. Which is not perfect solution, but it's works. There is a limitation on re-exporting modules with `export = ` (see https://github.com/Microsoft/TypeScript/issues/4336).

Tests should be exacly like in `typescript-definition-test.ts` but without PaperScript staff I suppose, and sadly with only default import. I think.

If you think this solution is worth it, I will add tests.

anion155:
Also, can re-export all named exports manually, like:
```typescript
export {
  Color: paper.Color,
};
```

sasensi:
I think that the best way to figure out if this is needed would be to set up a project reproducing the issue (using paper-core in a TypeScript project but losing typings) and see if this is the best way to solve it or if there is another workaround.
This [StackBlitz](https://stackblitz.com/edit/typescript-paperjs) can serve as a starting point.
Do you think that you could do that (maybe based on your own use case) ?

anion155:
Ok, [here it is](https://stackblitz.com/edit/typescript-paperjs-6xsbfu?file=index.ts). As you can see there is no typings for core version.

anion155:
@sasensi I did found out if I add to `paper.d.ts` declaring of module `paper/dist/paper-core` with `export = paper;`, and in file `paper/dist/paper-core.d.ts` just do `import 'paper';` everything work great.

sasensi:
@anion155, thank you for the reproduction case, it is much clearer for me now.
So yes, I would say that this is a valid use case and we should be able to find a solution for it.

Your last solution seems cleaner than your first proposal, but indeed, `PaperScript` class should not be available in `paper-core` module.
I guess that we could split the typescript declaration into two parts: the `paper-core` base on one side and the `paper-full` extra features on the other (I only see `PaperScript` class here for the moment).
Then, we would have 2 different module exports, as you propose.

This is the first time that I face this pattern (multiple possible exports of a library) so I don't know yet how to achieve that.
But I guess that we should be able to find another library that uses this pattern and see how they handle it.

sasensi:
@anion155, I think that I found a solution based on my idea of splitting the declaration in 2 parts.
Here is the [StackBlitz](https://stackblitz.com/edit/typescript-paperjs-solution?file=paper.d.ts&view=editor).
I think that it resolves your issue and it should still work in normal cases.
The only thing is that the declaration generation needs to be completely refactored but that's something that I can handle (as I wrote it) if the solution seems good to you too ?

anion155:
Just a tip. Do not forget to actually add file `paper-core.d.ts`, 'cause if you actually core module without reference-thing, or import full version somewhere, you would get error `Cannot find declarations for ...`.

sasensi:
@anion155, indeed, thanks for the tip.
I made a PR (#1716 ) that should solve this issue.
Please have a look at it and confirm that it works in your case too.

PR #1716, opened 2019-10-06
Allow paper core import in TypeScript

sasensi:
### Description
Typings were missing when importing paper core version with: 
`import * as paper from 'paper/dist/paper-core'` syntax.
This changes the generated TypeScript definition so that it exports two modules: `paper` and `paper/dist/paper-core`. In the same logic, `paper-core.d.ts` file is added to make sure that the corresponding definition is automatically loaded.
This also takes care of the fact that `PaperScript` class is not available in paper core version, by removing it from the corresponding TypeScript definition.
Finally, this also simplifies existing definition by directly exporting a `PaperScope` instance as the module instead of duplicating all `PaperScope` properties and methods on the module itself.

#### Related issues
- Closes #1713

### Checklist


dist/paper-core.d.ts
```diff
@@ -0,0 +1 @@
+import './paper';
```
anion155: https://eslint.org/docs/rules/eol-last

sasensi: Indeed, I didn't notice that yet but the rest of the files seems to follow this rule (even if not reported by the linter task).
I corrected that in the next commit.

gulp/typescript/typescript-definition-generator.js
```diff
@@ -92,20 +92,11 @@ classes.forEach(cls => {
...
-paperScopeClass.classesPointers = classes.filter(_ => _.className !== 'PaperScope').map(_ => ({ name: _.className }));
-
-// Since paper.js module is at the same time a PaperScope instance, we need to
-// duplicate PaperScope instance properties and methods in the module scope.
-// For that, we expose a special variable to the template.
-const paperInstance = { ...paperScopeClass };
-// We filter static properties and methods for module scope.
-paperInstance.properties = paperInstance.properties.filter(_ => !_.static);
-paperInstance.methods = paperInstance.methods.filter(_ => !_.static && _.name !== 'constructor');
+paperScopeClass.classesPointers = classes.map(_ => ({ name: _.className }));
```
lehni: Not a fan of `_` for the item of each iteration. Can we call it `it` instead , in both of the above lines?

sasensi: Oops, yes, I remember that we talked about that in the past :)

PR #1733, opened 2019-11-09
Fix handling of negative Shape.Rectangle sizes

### LINKED ISSUE #1245: Shape.Rectangle(point, size) draws Circle in some cases
If we use ```Shape.Rectangle(point, size)``` with ```size = new Size(x, y);``` where both ```x``` and ```y``` are negative, a **Circle is drawn** instead of a Rectangle. 

Example: 
```js
var point = new Point(50, 50);
var size = new Size(-10, -10);
var shape = new Shape.Rectangle(point, size);
shape.strokeColor = 'black';
```

Could be checked here: 
http://paperjs.org/reference/shape/#shape-rectangle-point-size

adroitwhiz:
### Description

This PR fixes the handling of negative Shape.Rectangle sizes by taking the absolute value of the shape's `size` to calculate the shape's new `radius` in `Shape.setSize`.

#### Related issues

Resolves #1245

### Checklist

### DIFF
--- a/src/item/Shape.js
+++ b/src/item/Shape.js
@@ -90,7 +90,7 @@ var Shape = Item.extend(/** @lends Shape# */{
                 height = size.height;
             if (type === 'rectangle') {
                 // Shrink radius accordingly
-                this._radius.set(Size.min(this._radius, size.divide(2)));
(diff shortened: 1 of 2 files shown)


lehni:
Great, thank you for the flawless PR!

PR #1738, opened 2019-11-29
Fix isEmpty return type erro in dist/paper.d.ts

NearZXH:


sasensi:
Hi, thank you for your contribution.
This issue has already been fixed with this commit: https://github.com/paperjs/paper.js/commit/cc15696750035ab00e00c64c7c95daa2c85efe01
By the way, the TypeScript definition is automatically generated from the code documentation so the way to fix TypeScript definition issues is not by updating the definition itself but rather by changing the corresponding documentation.

NearZXH:
That's great. Make me learn more.

Issue #1737, opened 2019-11-22
Use sweep-and-prune algorithm for overlap detection


waruyama:
Currently if overlaps between several curves or paths have to be found, we compare the bounds of each element to the bounds of all other elements. This works, but since it it a classic O(n2) algorithm, it becomes slow if many elements are involved.

A much better algorithm is Sweep and Prune, which sorts the elements along one axis and then sweeps along the axis finding potential overlaps. This is a O(n) algorithm, which means it has much better performance for many elements. Here is a great explanation on how it works:

https://gamedev.stackexchange.com/questions/57470/are-collision-detection-always-on2

Instance where Sweep and Prune could be used are:
- [Curve.getIntersections()](https://github.com/paperjs/paper.js/blob/15e00e0b99b5f59215028826a39321248f433d7a/src/path/Curve.js#L2119)
- [PathItem.compare()](https://github.com/paperjs/paper.js/blob/bba70907e79af3e990341493eb2aaf8c4f700f25/src/path/PathItem.js#L725)
- [PathItem.Boolean.reorientPaths()](https://github.com/paperjs/paper.js/blob/bba70907e79af3e990341493eb2aaf8c4f700f25/src/path/PathItem.Boolean.js#L306)

I had implemented this in an older version of Paper.js, but porting it to a current version should not be a problem.

For complex cases the performance improvement should be dramatic.

lehni:
This sounds really good. I can't wait for this to land!
BTW, you mean `Curve.getIntersections()`, right?

waruyama:
Thanks, corrected. This will only consist of two static functions:

`getItemBoundsOverlaps(items1, items2, tolerance)`

and

`getBoundsOverlaps(bounds1, bounds2, tolerance)`

The first is only a utilty function that takes the bounds of all items can calls the second function.
The return value will be an array. Each entry contains the indexes of  overlapping `items2`/`bounds2` entries for the entry at this index in `items1`/`bounds1`.

Any suggestions where to put these two functions?

lehni:
Do they both use the same algorithm internally? Could the algorithm itself be in `Numerical`? And then maybe two static methods that make use of it: `Rectangle. getBoundsOverlaps()` and `Items. getBoundsOverlaps()`?

lehni:
If only `Rectangle.getBoundsOverlaps()` uses the algorithm then it can be inlined there.

waruyama:
[...]

PR #1739, opened 2019-11-29
First implementation of sweep and prune

waruyama:
Implements a sweep and prune algorithm for collision testing of bounds.

I put the algorithm in a new file `CollisionDetection.js`, but this should be reconsidered. The reason why I did not put it into `Rectangle.js` is that it currently does not accept an array of Rectangles, but and array of coordinate arrays [x0, y0, x1, y1].

Anyway, this implementation passes all tests and is a starting point.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 /dist/*/
 /dist/*.min.js
 /dist/paperjs.zip
+package-lock.json
+package.json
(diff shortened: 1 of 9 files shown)


waruyama:
I will try again later with a new pull request only including the relevant files.

lehni:
You don't need to close the PR. You can add a fixing commit, even squash these commits to one, and then force-push the new branch to the one that you created the PR against, and it will update.

lehni:
Relates to #1737

PR #1740, opened 2019-11-29
First implementation of sweep and prune

waruyama:
Implements a sweep and prune algorithm for collision testing of bounds.

I put the algorithm in a new file CollisionDetection.js, but this should be reconsidered. The reason why I did not put it into Rectangle.js is that it currently does not accept an array of Rectangles, but and array of coordinate arrays [x0, y0, x1, y1].

Anyway, this implementation passes all tests and is a starting point.

Relates to #1737

### DIFF
--- a/src/path/PathItem.Boolean.js
+++ b/src/path/PathItem.Boolean.js
@@ -156,19 +156,61 @@ PathItem.inject(new function() {
             collect(paths1);
             if (paths2)
                 collect(paths2);
+
+            var curvesValues = new Array(curves.length);
+            for (var i = 0, l = curves.length; i < l; i++) {
+                curvesValues[i] = curves[i].getValues();
+            }
+            var horCurveCollisions = 
+                CollisionDetection.findCurveBoundsCollisions(
+                    curvesValues, curvesValues, 0, false, true);
+            var horCurvesMap = {};
+            for (var i = 0; i < curves.length; i++) {
+                var curve = curves[i];
+                var collidingCurves = [];                  
+                var collisionIndexes = horCurveCollisions[i];
+                if (collisionIndexes) {
+                    for (var j = 0; j < collisionIndexes.length; j++) {
+                        collidingCurves.push(curves[collisionIndexes[j]]);
+                    }
+                }
+                var pathId = curve.getPath().getId();
+                horCurvesMap[pathId] = horCurvesMap[pathId] || {};
+                horCurvesMap[pathId][curve.getIndex()] = collidingCurves;
+            }
+
+            var vertCurveCollisions = 
+                CollisionDetection.findCurveBoundsCollisions(
+                    curvesValues, curvesValues, 0, true, true);
+            var vertCurvesMap = {};
+            for (var i = 0; i < curves.length; i++) {
+                var curve = curves[i];
+                var collidingCurves = [];                  
+                var collisionIndexes = vertCurveCollisions[i];
+                if (collisionIndexes) {
+                    for (var j = 0; j < collisionIndexes.length; j++) {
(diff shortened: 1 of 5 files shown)


waruyama:
Here is a simple html page for testing. On my machine this runs about 3-4 times faster with the sweep and prune algorithm. Actually, I was expecting it to be even faster, but it is still a considerable improvement.

```html
<html>
  <head>
    <title>Performance test for sweep and prune</title>
    <script src="paper.js"></script>
    <script type="text/paperscript" canvas="myCanvas">
      var path1 = new CompoundPath();
      for (var i = 0; i < 40; i++) {
          for (var j = 0; j < 40; j++) {
              path1.addChild(new Path.Circle({center: [i * 25, j * 25], radius: 10}))
          }
      }
      var path2 = path1.clone().translate(5, 5);
      
      var starttime = Date.now();    
      var result = path1.intersect(path2);
      console.log(Date.now() - starttime);

      result.fillColor = "red"
    </script>
  </head>
  <body>
    <canvas id="myCanvas" width=1024 height=1024></canvas>    
  </body>
</html>
```

src/util/CollisionDetection.js
```diff
@@ -0,0 +1,303 @@
...
+ * All rights reserved.
+ */
+
+/**
+ * @name CollisionDetection
+ * @namespace
+ * @private
+ */
+var CollisionDetection = new (function() {
+    return /** @lends CollisionDetection */ {
```
lehni: Since you don't actually have anything private inside the scope, you can also just use a normal object literal here:

```js
var CollisionDetection = {
    findItemBoundsCollisions: function() { …. }
};
```

src/path/PathItem.Boolean.js
```diff
@@ -801,7 +854,12 @@ PathItem.inject(new function() {
                             }
                         }
                     }
-                    wind = wind || getWinding(pt, curves, dir, true);
+                    var pathId = path.getId();
+                    var curveIndex = curve.getIndex();
+                    var hCollisions = horCurveCollisionsMap[pathId][curveIndex];
+                    var vCollisions = vertCurveCollisionsMap[pathId][curveIndex];
```
lehni: Nitpick, but: Elsewhere we join such variable definitions with one `var`

src/path/PathItem.Boolean.js
```diff
@@ -300,29 +341,39 @@ PathItem.inject(new function() {
...
-                    // on winding, we have to find one containing path with
-                    // different "insideness" and set opposite orientation.
-                    if (path2.contains(point)) {
-                        var entry2 = lookup[path2._id];
-                        containerWinding = entry2.winding;
-                        entry1.winding += containerWinding;
-                        entry1.container = entry2.exclude ? entry2.container
-                                : path2;
-                        break;
+                    collisionsI = collisions[i];
```
lehni: Maybe `indices` instead of `collisionsI`?

src/util/CollisionDetection.js
waruyama: OK, will do.

src/path/PathItem.Boolean.js
waruyama: Jut be as nitpicking as possible, I want this to look as the rest of the paper code.

src/path/PathItem.Boolean.js
waruyama: Changed this. I also changed `indexes` to `indices` at another location in the code.

waruyama:
I have done some cleanup of the code and added more comments. In my opinion this is now "OK". Not sure how to proceeed now. Do you want any explanations what is happening in this new code?

PR #1742, opened 2019-12-05
Fix parsing of rgb color string with percentages

### LINKED ISSUE #1736: Rgb color with percent parsed incorrectly

`Color` incorrectly parses colors given with percentages. This manifest itself at least when SVGs with such colors are being parsed, but perhaps in other scenarios as well

a JS minimal working example:
```
var paper = require('paper-jsdom');

red = new paper.Color('rgb(100%,0%,0%)')
console.log(red.toString())
```
which prints
```
{ red: 0.39216, green: 0, blue: 0 }
```

# Expected result

```
{ red: 1, green: 0, blue: 0 }
```

# Additional information

I suspect the problem is in the code that parses the colors, which seems to assume `rgb` always has values between 0 and 255, see [here](https://github.com/paperjs/paper.js/blob/e5d7bafd39db546e6c78bdbd89212c1685e05e1b/src/style/Color.js#L104) (although the hsl code seems to be able to work with `%` just fine. Indeed, `rgb(255%,0%,0%)` produces `{ red: 1, green: 0, blue: 0 }`. Also, there's no test case for this syntax.

That this is a valid way of specifying color in SVGs can be seen in [the docs](https://www.w3.org/TR/SVGColor12/), see the example in section 3.1 and the syntax in section 9. Additionally, I've encountered this on real-life SVG data.

[It also seems to be a valid CSS color spec](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value). Search for "RGB syntax variations" or Functional Notation.

MacOS 10.14.6, Node v12.11.1

PS: I'll work around this for now by pre-processing my files. Also, I am not a JS person and do not have the project set up; but if there's enough will and interaction, I might dedicate some time to setting up the project and to fixing this myself.

sapics:
This PR fixes #1736

### DIFF
--- a/src/style/Color.js
+++ b/src/style/Color.js
@@ -102,7 +102,7 @@ var Color = Base.extend(new function() {
                     }
                 } else if (i < 3) {
                     // RGB color values to 0..1
-                    value /= 255;
+                    value /= /%$/.test(component) ? 100 : 255;
                 }
                 components[i] = value;
             }
--- a/test/tests/Color.js
+++ b/test/tests/Color.js
@@ -84,6 +84,9 @@ test('Creating Colors', function() {
     equals(new Color('rgba( 255, 0, 0, 0.5 )'), new Color(1, 0, 0, 0.5),
             'Color from rgba() string 2nd test');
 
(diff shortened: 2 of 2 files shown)


lehni:
Thank you @sapics!

PR #1763, opened 2020-01-21
Fix: closed Path with blend mode throw error

### LINKED ISSUE #1755: Path.removeSegments() results in script error in function addJoin (segment is undefined)
quad = new Path ({strokeWidth:2, strokeColor: "red", blendMode: "screen"});
quad.add ([100, 100]);
quad.add ([200, 100]);
quad.add ([200, 200]);
quad.add ([100, 200]);
quad.closed = true;
//quad.closed = false;
quad.removeSegments ();

This results in a script error when the blendMode is defined as anything other than "normal".

# Expected result
All segments should be removed from the Path, while it still retains its original properties.
Due to the error, however,  the script terminates execution.

Workarounds:
1) Set blendMode to "normal", or don't set it at all;

2) Before calling Path.removeSegments(), unclose the object by setting Path.close = false;

3) Wrapping the addJoin function in a conditional wrapper, e.g. if (typeof segment != "undefined")

# Additional information
Windows 10 Home, Firefox (72.0.1 64-bit), paperjs 11.3 - 12.4 + prebuilt,

sasensi:
### Description
In some specific case, closed path was assumed to have at least one segment whereas this not always the case.

#### Related issues

- Closes #1755

### Checklist

### DIFF
--- a/src/path/Path.js
+++ b/src/path/Path.js
@@ -2793,15 +2793,18 @@ statics: {
         }
 
         var length = segments.length - (closed ? 0 : 1);
-        for (var i = 1; i < length; i++)
-            addJoin(segments[i], join);
-        if (closed) {
-            // Go back to the beginning
-            addJoin(segments[0], join);
-        } else if (length > 0) {
-            // Handle caps on open paths
-            addCap(segments[0], cap);
-            addCap(segments[segments.length - 1], cap);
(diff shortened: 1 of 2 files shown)


lehni:
Thank you @sasensi!

PR #1764, opened 2020-01-21
Fix: SVG import error stops script execution

### LINKED ISSUE #1754: Strange behavior when importing SVG with id=number
#1204 Description/Steps to reproduce

I'm finding a bug where when an SVG is imported that contains an id that begins with a number, the onError fires but the SVG is still imported and added to the root of the DOM at the bottom.

Convert an SVG file to a paper.js object using the importSVG() method where the SVG contains <g id="2"...

The onError event handler should fire showing an error with the SVG file but the SVG should be added as a node of the root DOM element.

Note: if there is not error then there is not added DOM element and no problem.

# Expected result

onError event handler to fire and SVG to not be imported.

# Additional information

  macOS 10.15.2, Node 12.14.0

sasensi:
### Description
This wraps the `importer(...)` call in a `try catch` block to catch eventual import error and still allow DOM cleaning after import.  
Note that the error is still logged in the console so that the user can correct it.
This is the minimal solution that I could think of but maybe this would be better to include more of the code in the `try` block to prevent the same problem with other parts of the code ?  
Also, I didn't write the related test because I am not so sure about how we should/could test this kind of problem. @lehni, some idea ?

#### Related issues

- Closes #1754

### Checklist

### DIFF
--- a/src/svg/SvgImport.js
+++ b/src/svg/SvgImport.js
@@ -607,8 +607,17 @@ new function() {
             insertItems = settings.insertItems;
         settings.applyMatrix = false;
         settings.insertItems = false;
-        var importer = importers[type],
-            item = importer && importer(node, type, options, isRoot) || null;
+        var importer = importers[type];
+        var item = null;
+        if (importer) {
+            // Catch import errors to allow final clean-up.
+            try {
+                item = importer(node, type, options, isRoot);
+            } catch (e) {
+                console.error(e);
+            }
+        }
+
         settings.insertItems = insertItems;
         settings.applyMatrix = applyMatrix;
         if (item) {


src/svg/SvgImport.js
```diff
@@ -607,8 +607,17 @@ new function() {
             insertItems = settings.insertItems;
         settings.applyMatrix = false;
         settings.insertItems = false;
-        var importer = importers[type],
-            item = importer && importer(node, type, options, isRoot) || null;
+        var importer = importers[type];
+        var item = null;
```
sapics, line 611:
```diff
-        var importer = importers[type];
-        var item = null;
+        var importer = importers[type],
+            item;
```
For the style consistency, I like the later one.

Issue #1621, opened 2018-11-29
Consider supporting size argument in Raster constructor


lehni:
Wouldn't it be nice to be able to do this?

```js
const raster = new Raster({ width: 100, height: 200 })
raster.setPixel(...)
```

PR #1770, opened 2020-02-07
Fix Raster constructor

### LINKED ISSUE #1768: paper-jsdom-canvas Raster does not work with node-canvas elements
With `paper-jsdom-canvas@0.11.5` and `canvas@1.6.7`, I was able to do something like this on a Express server:

```
const { createCanvas, loadImage } from "node-canvas"
const Paper = require("paper-jsdom-canvas")

router.get("/test", (req, res, next) => {
  const ps = new Paper.PaperScope()
  ps.setup([1024, 1024])

  loadImage(
    "http://jewlrassets.s3-website-us-west-2.amazonaws.com/imgsmbly/JWL0013/t2/sku/w.png"
  )
    .then(img => {
      console.log("LOADED IMAGE", img)
      const canvas = createCanvas(img.width, img.height)
      const ctx = canvas.getContext("2d")
      ctx.drawImage(img, 0, 0)

      ps.activate()
      const r = new ps.Raster(canvas.toDataURL(), ps.project.view.center)
      r.onError = err => {
        console.log("ERROR:", err)
      }
      r.onLoad = () => {
        ps.view.draw()

        const quality = 0.9
        var jpeg = ps.view.element.jpegStream({
          quality: quality,
          progressive: true
        })
        res.set("Content-Type", "image/jpeg")
        jpeg.pipe(res)
      }
    })
    .catch(err => {
      console.log("ERROR", err)
    })
})
```

Now with the latest `0.12.4` and `canvas@2.4.1`, it results in the canvas not loading into the project. It's just a black image.

I have verified that the issue lies somewhere with Paper.js since streaming the canvas directly yields the correct image:

```
const { createCanvas, loadImage } from "node-canvas"
const Paper = require("paper-jsdom-canvas")

router.get("/test", (req, res, next) => {
  const ps = new Paper.PaperScope()
  ps.setup([1024, 1024])

  loadImage(
    "http://jewlrassets.s3-website-us-west-2.amazonaws.com/imgsmbly/JWL0013/t2/sku/w.png"
  )
    .then(img => {
      const canvas = createCanvas(img.width, img.height)
      const ctx = canvas.getContext("2d")
      ctx.drawImage(img, 0, 0)

      res.set("Content-Type", "image/jpeg")
      const stream = canvas.createJPEGStream()
      stream.pipe(res)
    })
    .catch(err => {
      console.log("ERROR", err)
    })
})
```

My use-case for using the canvas is because I need to load files directly from the server filesystem, but I used a URL here so you can have a reproducible test case.

sasensi:
### Description
Corrects `new Raster(size, position)` tests that lead to false positives and adds new tests for `new Raster(source, position)`.

#### Related issues

- Closes #1768

### Checklist

### DIFF
--- a/src/item/Raster.js
+++ b/src/item/Raster.js
@@ -112,28 +112,26 @@ var Raster = Item.extend(/** @lends Raster# */{
         // - An image (Image|Canvas|String) and an optional position (Point).
         // - A size (Size) describing the canvas that will be  created and an
         //   optional position (Point).
-        // If _initialize can set properties through object literal, we're done.
-        // Otherwise we need to check the type of object:       var image,
-        if (!this._initialize(source,
-                position !== undefined && Point.read(arguments))) {
-            var image,
-                type = typeof source,
-                object = type === 'string'
-                    ? document.getElementById(source)
-                    : type  === 'object'
-                        ? source
-                        : null;
-            if (object && object !== Item.NO_INSERT) {
-                if (object.getContext || object.naturalHeight != null) {
-                    image = object;
-                } else if (object) {
-                    // See if the arguments describe the raster size:
-                    var size = Size.read(arguments);
-                    if (!size.isZero()) {
-                        image = CanvasProvider.getCanvas(size);
-                    }
+        var image,
+            type = typeof source,
+            object = type === 'string'
+                ? document.getElementById(source)
+                : type === 'object'
+                    ? source
+                    : null;
+        if (object && object !== Item.NO_INSERT) {
+            if (object.getContext || object.naturalHeight != null) {
+                image = object;
+            } else if (object) {
+                // See if the arguments describe the raster size:
+                var size = Size.read(arguments);
+                if (!size.isZero()) {
+                    image = CanvasProvider.getCanvas(size);
                 }
             }
+        }
+        if (!this._initialize(source,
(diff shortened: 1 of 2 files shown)


lehni:
Thanks @sasensi for the fix! But while this indeed does fix the issue, it also makes the constructor slower since it has to do process a bunch of values that it might not even need, in case the call `_initialize()` returns true.

I therefore wonder:  Why was `new Raster(size, position)` introduced in the first place? Do we really need this constructor? you added it here: https://github.com/paperjs/paper.js/commit/314390d78911580a0bb43463598c9bbb881b4a9e

sasensi:
@lehni, yes, I understand your concern, as stated in the commit message, I remember adding it following your idea here: https://github.com/paperjs/paper.js/issues/1621
Maybe this was coming from a workshop that you did or something ?

lehni:
Ah yes, that's what it was :) The thing is this: The 2nd argument was only meant for internal use, and wasn't supposed to be exposed… I didn't notice it back then.

sasensi:
@lehni, in fact, as shown by https://github.com/paperjs/paper.js/issues/1854, even the original `new Raster(source, position)` is broken so I think that this fix is needed regardless of whether `new Raster(size, position)` should be exposed or not.

PR #1782, opened 2020-03-02
Update raster initialize docs

ChrisZieba:
I notice the Raster `initialize` method takes an object for the first parameter. The typescript definitions don't seem to have this overload. I also remove part of the function description that I believe is redundant (not too sure)? I couldn't find much documentation on how to update the typescript definitions, but it seems like they are auto-generated from the source.

### DIFF
--- a/src/item/Raster.js
+++ b/src/item/Raster.js
@@ -45,8 +45,8 @@ var Raster = Item.extend(/** @lends Raster# */{
      * describing all the properties to set.
      *
      * @name Raster#initialize
-     * @param {HTMLImageElement|HTMLCanvasElement|String|Object} [source] the source of
-     *     the raster
+     * @param {HTMLImageElement|HTMLCanvasElement|String|Object|Size} [source] the
+     *     source of the raster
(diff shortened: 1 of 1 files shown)


lehni:
Thank you @ChrisZieba for spotting this. But this is not how I'd like to fix this. Merging the docs to one is not the way to go. Instead, we should make the typescript generator respect multiple definitions of the same constructor. These constructor versions do very different things, and we keep them separate on purpose, see the reference:

http://paperjs.org/reference/raster/#raster-size

Also, we have the same situation in many other places...

http://paperjs.org/reference/path/#path-object

@sasensi could you take a look at how to add how to adjust the typescript typings generator?

sasensi:
@lehni, I think that the simplest way to handle this, is to have a dedicated entry in the documentation and the typescript definition for the constructor with object argument signature, like we do for `Path` for example. Here is the PR doing this: https://github.com/paperjs/paper.js/pull/1815
A second step, as we already discussed in other issues, would be to improve our JSDoc to be able to handle complete object descriptions.

PR #1796, opened 2020-04-06
Fix typescript returned instance type

### LINKED ISSUE #1795: path.clone() doesn't have the right return type in TypeScript

```typescript
import * as paper from 'paper';

let path = new paper.Path();
path = path.clone();
```

The last assignment produces an error because `path.clone()` return type is `Item` and not `Path` as it should be.

# Expected result

This code should not produce any error.

# Workaround

```typescript
import * as paper from 'paper';

let path = new paper.Path();
path = path.clone() as paper.Path;
```

sasensi:
### Description

See the related issue for a reproduction case.  
The problem is that JSDoc itself doesn't have a way to solve this issue, as shown here: https://github.com/jsdoc/jsdoc/issues/991  
So the best workaround (also used by other libraries) seems to be using a custom JSDoc tag that is only understood by the TypeScript definition generator. When this tag is detected, the return type is replaced by TypeScript `this` type which perfectly solves this issue (it is actually done for this specific use case).

#### Related issues

- Closes #1795

### Checklist

### DIFF
--- a/gulp/typescript/typescript-definition-generator.js
+++ b/gulp/typescript/typescript-definition-generator.js
@@ -329,10 +329,10 @@ function formatJSDoc(offset, render) {
 function getMethodReturnType(method) {
     // We need to handle the special case where the method returns an instance
     // of the class. As JSDoc doesn't have a proper way to handle this, we rely
-    // on the custom tag `@tsReturnThis`, placed in the description of the
+    // on the custom tag `@chainable`, placed in the description of the
     // method. When this tag is present, we ignore the common return type and
     // use typescript `this` type instead.
-    if (method.comment.tags.find(it => it.title === 'tsReturnThis')) {
+    if (method.comment.tags.find(it => it.title === 'chainable')) {
         return 'this';
     }
     return method.returnType || method.returns.length > 0 && method.returns[0].type;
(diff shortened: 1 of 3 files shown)


gulp/typescript/typescript-definition-generator.js
```diff
@@ -327,6 +327,15 @@ function formatJSDoc(offset, render) {
...
 function getMethodReturnType(method) {
+    // We need to handle the special case where the method returns an instance
+    // of the class. As JSDoc doesn't have a proper way to handle this, we rely
+    // on the custom tag `@tsReturnThis`, placed in the description of the
+    // method. When this tag is present, we ignore the common return type and
+    // use typescript `this` type instead.
+    if (method.comment.tags.find(it => it.title === 'tsReturnThis'))
+    {
+        return 'this';
+    }
```
lehni, line 338: Let's adhere to Paper.js code-style:

```js
if (method.comment.tags.find(it => it.title === 'tsReturnThis')) {
    return 'this';
}
```

sasensi: Oops, indeed.

sasensi:
> This is great! I would suggest we call the tag simply `@chainable` though, as it's not specific to TS anyway? (See: [jsdoc/jsdoc#991 (comment)](https://github.com/jsdoc/jsdoc/issues/991#issuecomment-96966299))

No, this was a totally made up name, so we can totally change it (which I just did in the last commit).

lehni:
Great, thanks!

PR #1797, opened 2020-04-09
docs: Fix simple typo, horziontal -> horizontal

timgates42:
There is a small typo in src/item/Item.js, src/view/View.js.

Should read `horizontal` rather than `horziontal`.


sapics:
Thanks @timgates42 !

PR #1800, opened 2020-04-13
Fix travis build configuration warnings

sapics:
This PR just fix warnings which are reported in
https://travis-ci.org/github/paperjs/paper.js/builds/674265980/config

I keep the configuration without `dist` key, because the `default` looks good for this project.


lehni:
Thank you @sapics!

Issue #1354, opened 2017-07-04
Support w3c pointer events api


allforabit:
It would be nice to have support for the Pointer events api: https://www.w3.org/TR/pointerevents/ With this it's possible to detect which device is being used for input, e.g. a stylus. Most browsers not support it apart from Safari, however there is a polyfill to work around this: https://github.com/jquery/PEP

siddhanathan:
Pointer Events Level 2: https://w3c.github.io/pointerevents/

sasensi:
Could you elaborate more about the feature that you are thinking about (use cases, ideal API code examples, ...) ?

sberney:
Pointer Events is a well designed API that supports most if not all touch capabilities that can be implemented in a browser, as it was intentionally designed to replace Touch Events and Mouse Events. It allows me to differentiate between the eraser on my Surface Book's Stylus, and its nib. It should support Apple Pencil, but I haven't verified. It supports multi-touch. Here's [a relevant subsection](https://w3c.github.io/pointerevents/#pointerevent-interface) from ghost's link describing the following properties that Pointer Events makes available -- it has strictly more information:

```
    long        pointerId = 0;
    double      width = 1;
    double      height = 1;
    float       pressure = 0;
    float       tangentialPressure = 0;
    long        tiltX = 0;
    long        tiltY = 0;
    long        twist = 0;
    DOMString   pointerType = "";
    boolean     isPrimary = false;
```

Pointer Events are implemented in [all the major relevant browsers](https://caniuse.com/#feat=pointer-events). Older browser versions are less compatible, so you would likely want to allow the user to opt into one or the other -- or gracefully fall back.

For my purposes, it doesn't work to only have information available from Touch Events, since I want to make a richly interactive program. I need to have that fine grain distinguishedness between all the different kinds of input buttons, and react accordingly. (And down the line, I can see it being nice to have access to those extra tilt and twist properties.) ... I bet a lot of other people would appreciate this too. I don't know what kinds of programs other people are building, but they presumably care about user input.

Issue #1179, opened 2016-10-02
onMouseDown prevents native onmousedown event from firing in IE


Drahoslav7:
In internet explorer 11, if there is tool with ```onMouseDown``` defined, then native ```onmousedown``` event on canvas does not work.
See https://codepen.io/anon/pen/JRABgG

I want to be able to detect right click on canvas outside of my paper.js drawing logic, but this seems to be impossible in IE.

kevinbrink:
Any updates on this?

Having a very similar issue with the `mousemove` event when tying `PaperJS` to [Leaflet](https://github.com/Leaflet/Leaflet).

Works fine in Chrome, and breaks to pieces on IE11.

lehni:
Here the same as [Sketch](http://sketch.paperjs.org/#S/bY+xDoJADIZfpbkFSAjsGDcdHJwcxeE8ilzAv4Y7YCC8u6BRo3Fok7Zf8v0dFfSVVaYONXtTqVgZKZa57GC8FZBgL53jjQwIuWf4iMYcREbgpOGkkUsYeJGGrgtXzFwQ04NMnn2orKmiVY5prhxGo9cuEbx5WtNL96VIU/KVdYR52VJpW3ZkQbvtcvxJAO1tz38yfOzzc+eWdX0TC+9UdjxNdw==)

lehni:
This looks like a IE bug. In your View handling code at the end of the mouse handling, we do the following:

            // Now call `preventDefault()`, if any of these conditions are met:
            // - If any of the handlers were called, except for mousemove events
            //   which need to call `event.preventDefault()` explicitly, or
            //   `return false;`.
            // - If this is a unhandled mousedown event, but the view or tools
            //   respond to mouseup.

Please note that `event.preventDefault()` is not the same as `event.stopPropagation()` and shouldn't prevent events from bubbling up. It should just prevent the default native behavior from happening, e.g. scrolling, dragging, etc.

But for some reasons, on IE, calling this has the same effect, and the event bubbling appears to be cancelled. This looks like a bug to me, and I'm not sure how we can circumvent that, without breaking a ton of other situations that this automatic calling of `event.preventDefault()` was created for.

You could of course install event handlers on the view the paper.js way, and those are handled separately from the tools, but are actually fired first:

[...]

PR #1810, opened 2020-05-19
Allow pointer events to be used if supported

### LINKED ISSUE #1463: Pen tablets don't register mouse events on Firefox 59 + Windows 10

- Use a Pen tablet. Any tablet that reports pen pressure and will trigger [PointerEvents][pe-docs].
- Visit this [Sketch][test-case].
- Try to draw with the pen tablet.

http://sketch.paperjs.org/#S/dY/BCoMwDIZfJfSigvgAhZ2262Cwo/VQa6ZSTaSt8yC++6yMDQbL6c/3J+TPKkiPKKS4WwymE7kw3MT+qR1MOnSKFEXNdOXZ44UXghM8ZjKhZ0oxg1URHKM7J1zgtsv0gLE8tiNS8BJKLCbuKVT5xwuOLZ55YCchqQdtbPI1cUATsJEQ3IyRbpmi7TeP0+2/PIVumvR9NK7uz9UOtT2AF7Ksthc=

# Expected result

- Draws freehand drawings

# Actual result

- Draws nothing

# Additional information

Only applicable on Windows 10 + Firefox 59.

[pe-docs]: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events
[test-case]: http://sketch.paperjs.org/#S/dY/BCoMwDIZfJfSigvgAhZ2262Cwo/VQa6ZSTaSt8yC++6yMDQbL6c/3J+TPKkiPKKS4WwymE7kw3MT+qR1MOnSKFEXNdOXZ44UXghM8ZjKhZ0oxg1URHKM7J1zgtsv0gLE8tiNS8BJKLCbuKVT5xwuOLZ55YCchqQdtbPI1cUATsJEQ3IyRbpmi7TeP0+2/PIVumvR9NK7uz9UOtT2AF7Ksthc=

adroitwhiz:
### Description
This PR adds support for using pointer events instead of mouse/touch events if supported in the browser.

Previously, paper.js was testing for pointer event support by checking `navigator.pointerEnabled`, but that was [removed from the pointer events spec back in 2013](https://webplatform.github.io/docs/dom/Navigator/pointerEnabled/) and no longer implemented in modern browsers. Now it tests for the existence of `window.PointerEvent`.

#### Related issues

- Relates to #1354
- Resolves #1463

### Checklist

  - This kind of thing is next to impossible to test automatically...


src/view/View.js
```diff
@@ -1070,7 +1072,7 @@ new function() { // Injection scope for event handling on the browser
     // Touch handling inspired by Hammer.js
     var navigator = window.navigator,
         mousedown, mousemove, mouseup;
-    if (navigator.pointerEnabled || navigator.msPointerEnabled) {
+    if (window.PointerEvent || navigator.msPointerEnabled) {
```
lehni, line 1075: I think it should now be: `window.PointerEvent || navigator.MSPointerEvent`, see: https://github.com/hammerjs/hammer.js/blob/563b5b1e4bfbb5796798dd286cd57b7c56f1eb9e/src/input/pointerevent.js#L34

adroitwhiz:
One other thing: Pointer events seem to break UI code that expects mouse events. Should pointer events have to be explicitly enabled, or should there be a version bump of some sort, to indicate that this may change applications' behavior?

lehni:
@adroitwhiz could you explain how they break things? I don't quite understand the problem

adroitwhiz:
If there's external code that expects you to *click* on a canvas (e.g. an eyedropper tool, or dropdown that is supposed to close when you click on another element), for some reason it won't receive those "click" events if pointer events are enabled. I ran into this when testing out this change on a real-world [codebase](https://github.com/llk/scratch-paint).

adroitwhiz:
Looks like the "pointer events break mouse events" behavior was talked about a bit [here](https://github.com/paperjs/paper.js/issues/1179#issuecomment-273645978).

danrha:
Would this potentially fix https://github.com/paperjs/paper.js/issues/1959?

eeropic:
@danrha yes I think so. I'm doing this locally for easier differentiation between stylus, touch and mouse inputs at the moment by changing `if (navigator.pointerEnabled || navigator.msPointerEnabled) {` 
to `if ('PointerEvent' in window) {`

mydudechris:
What is the status with this PR @lehni? We've run into an issue that requires pointer support, a drawing app for use with iPad and Wacom tablets that are listening for pointer events (eyedropper, drawing, etc). Any chance we can get this pushed?

zalo:
I've found that the `window.PointerEvent` check doesn't work in modern Chrome on Windows, but `"onpointerdown" in window` appears to work well.

I've needed this to get the `pointerId` to enable [setPointerCapture](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture).

PR #1818, opened 2020-06-03
Add JSDoc comments for additional PaperScope methods

lazerwalker:
### Description

The TypeScript type definitions don't include `createCanvas` or `clear` on the PaperScope object. This PR adds JSDoc comments for those two functions so they show up in both the online documentation and the generated `.d.ts` typings files.

This is a comment/documentation-only change.

### Checklist

### DIFF
--- a/src/core/PaperScope.js
+++ b/src/core/PaperScope.js
@@ -265,6 +265,19 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
         return this;
     },
 
+    /**
+     * Returns a valid HTML canvas you can use. Internally, this may reuse an existing canvas.
+     *
+     * This either accepts width and height as two arguments, or you can pass in an object with "width" and "height" properties.
+     *
+     * @param {Number|Object} width
+     * @param {Number} height
+     *
+     * @option width {Number}
+     * @option height {Number}
+     *
+     * @returns {HTMLCanvasElement}
+    */
     createCanvas: function(width, height) {
         return CanvasProvider.getCanvas(width, height);
     },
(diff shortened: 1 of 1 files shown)


lehni:
Great, thank you @lazerwalker! Could you make it wrap at 80 chars? All code does in paper.js

And I have a feeling there is a better way to document the two separate version of the method signature. @sasensi since TS is your area, what do you think? I think two separate declarations would be better, yes?

sasensi:
@lehni, yes indeed, I think that this would be more consistent with the rest of the codebase.

sasensi:
@lazerwalker, indeed, we don't support yet deep object description so `Object` type is the best we can provide for now.  
About splitting the documentation into 2 signatures, here is an example of how we do this:
https://github.com/paperjs/paper.js/blob/b081fd72c72cd61331313c3961edb48f3dfaffbd/src/item/Raster.js#L75-L117

lazerwalker:
Just split it up into two blocks! Let me know if there are any other issues.

Also happy to squash into a single commit before merging if it's at that point + you'd prefer.

src/core/PaperScope.js
```diff
@@ -266,20 +266,26 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
...
      * @param {Number} height
      *
+     * @returns {HTMLCanvasElement}
+     */
+
+    /**
+     * Returns a valid HTML canvas you can use.
+     * Internally, this may reuse an existing canvas.
+     *
+     * @param {Object} options An object containing the canvas width / height
```
lehni, line 282: This should be `Size` really.

src/core/PaperScope.js
```diff
@@ -266,20 +266,26 @@ var PaperScope = Base.extend(/** @lends PaperScope# */{
...
+     *
      * @option width {Number}
      * @option height {Number}
```
lehni, line 285: When using @option, it should include the object param name as well, so `options.width` and `options.height` in your case. But since it should really be a Size object, this isn't even needed.

Issue #1824, opened 2020-06-11
strokeBounds is incorrect for certain paths


adroitwhiz:
# Description/Steps to reproduce

In some cases, paths' `strokeBounds` will be incorrectly sized too small.
This can easily be seen with e.g. a rotated ellipse, as in the sketch below.

1. Create a `paper.Path.Ellipse`
2. Rotate it
3. Get its stroke bounds
4. Notice that the bounds do not cover the actual stroke (and vary depending on the rotation)

# Link to reproduction test-case

[Sketch](http://sketch.paperjs.org/#V/0.12.7/S/XZBPC8IwDMW/SullE8aYiiAbXhTvogcPboeuC26spqONCIrf3WxD/NNDSn59ebz0IVFdQKby0ALpWkZS26rvtUVPAsRKINxEpzpw8U5RHW+NaToP4SNHwaezDVIqTtMkiQSXIhq5b+7AeMF0mRQ5PidZjjlC7MnZFo5NRTWbz2fZB26ssY5hUBql22B4cZYUQbhIxvkxVmmvWHlWvifXA/hSOND0k/1Q8xXvGSs8GwhHi960l/6lmv7ir1xXCDL+pNKBaofVvUxPxfMF)

# Expected result

An item's `strokeBounds` should cover the entire item-- otherwise, this can lead to viewboxes of exported SVGs being incorrectly shrunken and cutting off outlines.

# Additional information

This occurs across browsers/OSes

lehni:
@adroitwhiz yes that's a bug indeed! Your fix looks good. I'll do some testing shortly, but it looks good to go! Out of curiosity: Github says `LLK/paper.js` is 91 commits ahead. Is there other work that we should bring over?

adroitwhiz:
There's a lot of stuff related to text import/export, most of which (I think) is specific to the Scratch paint editor.

There's a couple other things that may be worth merging:
- https://github.com/LLK/paper.js/pull/5 - add an option to allow hit tests to hit paths with no fill (plus [a follow-up](https://github.com/LLK/paper.js/pull/9) and [to-be-merged addition that does the same thing for paths with no stroke](https://github.com/LLK/paper.js/pull/39))
- https://github.com/LLK/paper.js/pull/28 - fix style merging in SVG export; I'm working on a more complete version of this that I plan to submit here as well

@fsih Any other LLK-specific paper.js fixes that could be upstreamed here?

lehni:
[...]

PR #1825, opened 2020-06-14
 Fix stroke bounds for paths

adroitwhiz:
### Description
The padding-plus-minmax check in https://github.com/paperjs/paper.js/pull/1082 was being done backwards-- we should have been checking if values were less than the current minimum with padding subtracted from the *values*, and greater than the current maximum with padding added, but we were subtracting padding from the *minimum* (and adding padding to the maximum).

Instead, we now add padding to the minimum and subtract it from the maximum, which is equivalent to subtracting padding from the values when comparing them to the minimum and adding padding to the values when comparing them to the maximum.

#### Related issues

- Relates to #1824

### Checklist

### DIFF
--- a/src/path/Curve.js
+++ b/src/path/Curve.js
@@ -858,8 +858,8 @@ statics: /** @lends Curve */{
         }
 
         padding /= 2; // strokePadding is in width, not radius
-        var minPad = min[coord] - padding,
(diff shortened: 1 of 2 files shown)


lehni:
LGTM, thanks!

PR #1860, opened 2020-10-02
Add null check for 2d context in CanvasView

dobesv:
### Description

Sometimes we are getting errors logged and reported with the vague error message "null is not an object (evaluating r.save)".

I traced it to this line.

It seems there was already a similar check & error in CanvasProvider so I copied that error here.

Although this does not prevent the problem, it at least will give us a clearer error when the problem occurs.

As for why a 2D context cannot be created, we are still looking into it.  Maybe the device (iPad) is running out of graphics memory?


dobesv:
I believe I found the cause of the issue that led me to this: https://bugs.webkit.org/show_bug.cgi?id=195325

Just a side-note for anyone running into this issue.  This PR does not fix that issue, it just makes the error message clearer when it does happen.

PR #1865, opened 2020-11-03
Add: support for arguments reading pattern in typescript definition.

sasensi:
### Description
Introduces hardcoded generic types in the typescript definition template that
are used to support arguments reading pattern (that allows a method to accept
either a Point instance or an array of number or an object...).

Maybe this can find a place in JSDoc at some point but I think that this does the job for now.

#### Related issues

- Relates to https://github.com/paperjs/paper.js/issues/1659

### Checklist


HriBB:
@sasensi what is the status on this PR? I really need this ;) Typescript types are basically useless if you use string colors or array points. 

@lehni can we merge this?

lehni:
I'm so sorry for never looking into this. The code looks good! It only adds the versions that read from objects, while in JS you can also do `<METHOD_NAME>(x, y)` instead of `<METHOD_NAME>({ x, y })` where `pointLike` values are read. With this change, only `<METHOD_NAME>([x, y])` will be added.

PR #1936, opened 2021-07-21
fix: requestAnimationFrame should ensure bind window

yhx9527:
### Description

nativeRequest should be bind wnidow, otherwise an error may occur
### Checklist


lehni:
Did an error actually occur? I've never seen this happening, so am really curious in what environment this is a problem, and what the error is?

PR #1988, opened 2022-08-01
Fix createPath arguments not being passed down to Path constructor

HriBB:
### Description
I believe I have found a bug in `paper.js`. I am using javascript directly, so no PaperScript. My use case is the following:

```js
// insertItems is turned off globally because I am using react-reconciler to handle rendering
paper.settings.insertItems = false;

// one of my tools has a use case, where I need to bypass react's rendering mechanism
var circle = new paper.Path.Circle({ insert: true });

// but circle is not inserted into the project and thus not visible until I render it with react
console.log(circle.isInserted()) // false
```
`Path.Circle` calls `createEllipse`, which calls `createPath`, which does not pass arguments to Path constructor.

### Checklist

### DIFF
--- a/src/path/Path.Constructors.js
+++ b/src/path/Path.Constructors.js
@@ -22,7 +22,10 @@ Path.inject({ statics: new function() {
 
     function createPath(segments, closed, args) {
         var props = Base.getNamed(args),
-            path = new Path(props && props.insert == false && Item.NO_INSERT);
+            path = new Path(props && (props.insert == true
+                ? { insert: true }
+                : props.insert == false && Item.NO_INSERT
+            ));
         path._add(segments);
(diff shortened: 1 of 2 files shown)


src/path/Path.Constructors.js
```diff
@@ -22,7 +22,10 @@ Path.inject({ statics: new function() {
 
     function createPath(segments, closed, args) {
         var props = Base.getNamed(args),
-            path = new Path(props && props.insert == false && Item.NO_INSERT);
+            path = new Path(props && (props.insert == true
+                ? { insert: true }
+                : props.insert == false && Item.NO_INSERT
+            ));
```
lehni, line 28: I think we can simply pass on `props`, but will merge first and check after.

PR #2066, opened 2024-02-16
Adds support for letter spacing

kennethormandy:
### Description

This pull request adds support for letter spacing, via the growing support for the [canvas letter spacing property](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/letterSpacing), and the existing support in SVG.

- SVG `letter-spacing`: [~96% browser support](https://caniuse.com/mdn-svg_attributes_presentation_letter-spacing)
- Canvas `letterSpacing`: [~76% browser support](https://caniuse.com/mdn-api_canvasrenderingcontext2d_letterspacing) (not supported in Safari)

All browser tests are passing, including some new ones for this feature.

### Example

example

```js
var text = new PointText({
  fontFamily: 'Charter',
  fillColor: 'black',
  content: 'Paper.js',
  // Etc.

  letterSpacing: 10,
});
```

### API considerations

I’d be interested in getting your feedback on a couple of possible adjustments, @lehni:

- **Number only?** Should `letterSpacing` only accept a number that’s intended to be relative to the font size, like `0.2`, rather than a value with units, like `'10px'`? Ie. Treat all values like ems? This appears to be [what OpenType.js does](https://github.com/opentypejs/opentype.js/pull/219), and it feels like the better way to use it to me.
- **Tracking or Letter Spacing** Depending on that, should it actually be called `tracking`, and use integers? I tend to prefer the CSS-like names, but it feels a bit like `leading` and `justification` in that `tracking` would be the more expected name within Paper.js.

### Related issues

- None that I could find

### Checklist

***

Thanks for considering this, and for all the great work on this library over the years!

### DIFF
--- a/dist/paper.d.ts
+++ b/dist/paper.d.ts
@@ -6538,6 +6538,11 @@ declare namespace paper {
          */
         justification: string
 
+        /** 
+         * The letter spacing of text content.
+         */
+        letterSpacing: string
+
 
         /** 
          * Style objects don't need to be created directly. Just pass an object to
@@ -6656,6 +6661,11 @@ declare namespace paper {
          * The justification of text paragraphs.
          */
         justification: string
+
+        /** 
+         * The letter spacing of text content.
+         */
+        letterSpacing: string
 
 
     }
--- a/gulp/typescript/typescript-definition-test.ts
+++ b/gulp/typescript/typescript-definition-test.ts
@@ -892,6 +892,7 @@ style.fontWeight;
 style.fontSize;
 style.leading;
 style.justification;
+style.letterSpacing;
 
 
 //
@@ -977,6 +978,7 @@ textItem.fontWeight;
 textItem.fontSize;
 textItem.leading;
 textItem.justification;
+textItem.letterSpacing;
 
 
 //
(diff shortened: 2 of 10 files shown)


jtuffier:
Hi @kennethormandy , and thank you for working on this subject. I was waiting for this for the last 2 years.
I have integrated your changes in my Angular application where I use paperJS, and everything seems to work perfectly in development mode.
I have an issue when I package my application in Electron : the letter spacing does not work at all. I checked in my JS scripts and I clearly see that the letter-spacing is in these scripts as expected. Do you have any idea why it does not work in Electron ? 
I use Angular 15 and Electron 28.

kennethormandy:
@jtuffier Glad it was useful for you! My guess would be that the underlying version of Chromium, depending on what version of Electron you’re using, might not support the letter spacing feature yet. So it’s like you’re viewing it in an older browser. The links in the PR description should still be relevant for that.

jtuffier:
@kennethormandy You were right, I was using the latest version for development (v33.0.0), but once my application was packaged I was using Electron v8.2.4. After updating the Electron dependency to v33.0.0 for my packaged application ,it works like a charm. Many thanks for your great support.

